PR #9, opened 2020-10-13
Use quarkus.native.additional-build-args instead of additionalBuildArgs

zakkak:
additionalBuildArgs is deprecated (see https://github.com/quarkusio/quarkus/issues/12666#issuecomment-707596054)


Karm (approved):
thx

PR #10, opened 2020-10-23
Remove obsolete libffi-devel dependency

zakkak:
`libffi-devel` as of 20.1.0.3.Final is no longer needed (https://github.com/graalvm/mandrel/issues/158)


jerboaa (approved):
LGTM

PR #17, opened 2021-02-23
Add reproducer for Version parsing IAE

jerboaa:


Karm:
@jerboaa  https://github.com/jerboaa/mandrel-integration-tests/pull/1

jerboaa:
Thanks @Karm merged that in.

Karm:
@jerboaa  @zakkak So, I read the issue, and IIUC, this is actually supposed to keep failing until the version parsing logic is changed, right?  Because that is what it does for me...

zakkak:
Correct

jerboaa:
Yes, it will only pass once https://github.com/graalvm/mandrel/issues/218 is fixed.

PR #18, opened 2021-02-26
Implements static check for harfbuzz lib and others

### LINKED ISSUE #16: Write integration test which exercises libfontmanager.a
Similar to the ImageIO integration test work, but using some code which actually exercises libharfbuzz.so code too (which previously was included in libfontmanager.so and is a separate library on some systems - or a system library). One test that fits the bill is [RotatedFontTest.java from OpenJDK 11](https://hg.openjdk.java.net/jdk-updates/jdk11u-dev/file/4ed322bf6b00/test/jdk/java/awt/font/Rotate/RotatedFontTest.java). This would ensure that generating native images from fonts-using code won't regress.

Karm:
Closes #16 

The test works on Centos 8 and on Ubuntu 18.   (Mandrel 21.0)

It fails on Windows with notorious java.lang.UnsatisfiedLinkError: no awt in java.library.path as that is not yet solved. I will take a look at it upstream (awt.lib and awt.dll are there, so it might not be that much of a problem...).


zakkak:
> It fails on Windows with notorious java.lang.UnsatisfiedLinkError: no awt in java.library.path as that is not yet solved. I will take a look at it upstream (awt.lib and awt.dll are there, so it might not be that much of a problem...).

AFAIK AWT support is not yet implemented for Windows

Karm:
@zakkak Shamelessly hijacking this PR's comments sections:

What does it mean "implemented"?  I might be very naive...though it seems to me that native calls should be satisfied by their backing in the .dll and we have the .lib with symbols to link to....so....why wouldn't it work once paths, libraries naming and linker flags are all set? What would there be to develop?

jerboaa:
> @zakkak Shamelessly hijacking this PR's comments sections:
> 
> What does it mean "implemented"? I might be very naive...though it seems to me that native calls should be satisfied by their backing in the .dll and we have the .lib with symbols to link to....so....why wouldn't it work once paths, libraries naming and linker flags are all set? What would there be to develop?

Code paths through native AWT libraries are different, JNI functions are different depending on the platform. Some AWT native functions are Windows only. Thus, registrations for JNI need to be different in Graal VM. Upstream AWT support is [Linux only in Graal VM 21.0](https://www.graalvm.org/release-notes/21_0/#native-image). See linked release notes.

jerboaa (approved):
Patch seems OK to me as a first shot. Hopefully the order of static libs is somewhat predictable. But we can fix that as a follow-up if it turns out to be a problem.

zakkak (approved):
LGTM, thanks Karm

PR #19, opened 2021-03-09
System.getenv() behavior test, Linux as expected, Windows fails

Karm:
Origins: https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/Native.20image.20Windows.20and.20ProcessEnvironment

TL;DR:

SmallRye Config hit this discrepancy between Linux and Windows where with Linux on ```System.getenv()``` a plain [Map](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/unix/classes/java/lang/ProcessEnvironment.java#L90) is returned and with Windows, the whole `ProcessEnvironment extends HashMap<String,String>` is [returned](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/windows/classes/java/lang/ProcessEnvironment.java#L233).

I am still not sure if it's a bug or a feature. Here is a tiny reproducer test so as it's not forgotten on Quarkus Zulip as I find it definitely as interesting, seemingly unnecessary (Case insensitive Windows env props?)  platform difference.

## Linux

```
2021-03-09 15:13:59.765 INFO  [o.g.t.i.AppReproducersTest] (getEnvtests) Testing app: GETENV
2021-03-09 15:14:23.512 INFO  [o.g.t.i.AppReproducersTest] (getEnvtests) Running...
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 24.188 s - in org.graalvm.tests.integration.AppReproducersTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
```

## Windows
```
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 21.628 s <<< FAILURE! - in org.graalvm.tests.integration.AppReproducersTest
[ERROR] getEnvtests{TestInfo}  Time elapsed: 20.894 s  <<< FAILURE!
org.opentest4j.AssertionFailedError:
build.log log should not contain error or warning lines that are not whitelisted. See testsuite\target\archived-logs\org.graalvm.tests.integration.AppReproducersTest\getEnvtests\build.log and check these offending lines:
Error: No instances of java.lang.ProcessEnvironment are allowed in the image heap as this class should be initialized at image runtime. To see how this object got instantiated use --trace-object-instantiation=java.lang.ProcessEnvironment.

Error: Image build request failed with exit status 1
Error: Use -H:+ReportExceptionStackTraces to print stacktrace of underlying exception ==> expected: <true> but was: <false>
        at org.graalvm.tests.integration.AppReproducersTest.getEnvtests(AppReproducersTest.java:317)
```


Karm:
Case sensitiveness: https://mail.openjdk.java.net/pipermail/core-libs-dev/2021-March/075042.html

Karm:
Thinking about this again...I actually lost the confidence that it warrants a test here.  A test for what? That it would consistently pass on Linux and fail on Windows? I should probably make it do something else :confused:

jerboaa:
@Karm I concur. This seems an application (framework?) bug to me. Make ProcessEnvironment initialized at runtime (if need be). I'm not sure there is a bug in OpenJDK.

PR #53, opened 2021-08-03
Fix for non-existent ? (question mark) dir

Karm:


Karm:
GW is broken on JDK installation. Not out fault.

PR #61, opened 2021-10-28
CI yaml fixed after PR 60 changes, Whitelist regexp fix

Karm:


zakkak:
@Karm FYI https://github.com/quarkusio/quarkus/pull/20355 is marked for backport to Quarkus 2.2 so eventually it will be the same for both.

.github/workflows/local_tests.yml
```diff
@@ -22,12 +23,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        quarkus-version: ['2.2.3.Final', '2.4.0.Final']
-        mandrel-version: ['21.2.0.2-Final', '21.3.0.0-Final']
+        quarkus-version: ['1.11.7.Final', '2.2.3.Final', '2.4.0.Final']
```
zakkak, line 26: Are we still interested in Quarkus 1.11 and Mandrel 20.3?

Karm: @zakkak Not for testing Q or M. This is for knowing that TS is still compatible with it. It is too soon to go and purge all ```if (QUARKUS_VERSION.startsWith("1.")) {...``` and thus remove the compatibility.

Karm:
@zakkak Seems fine now. Ad perf results: I will examine them on a baremetal lab box.

PR #64, opened 2021-11-10
The artifact quarkus-vertx-web is now named quarkus-reactive-routes

zakkak:
Closes: #63

Note: A potential issue with this patch is that it won't work with snapshots from Quarkus versions before the artifact renaming.

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
@@ -255,6 +255,9 @@ public class DebugSymbolsTest {
             if (QUARKUS_VERSION.startsWith("2.4")) {
                 runCommand(getRunCommand("git", "apply", "quarkus_2.4.x.patch"),
                         Path.of(BASE_DIR, Apps.DEBUG_QUARKUS_BUILDER_IMAGE_VERTX.dir).toFile());
+            } else if (QUARKUS_VERSION.contains("SNAPSHOT")) {
+                runCommand(getRunCommand("git", "apply", "quarkus_main.patch"),
+                        Path.of(BASE_DIR, Apps.DEBUG_QUARKUS_BUILDER_IMAGE_VERTX.dir).toFile());
             }
 
             // Build & Run
--- a/builder-image-apps/quarkus-vertx/quarkus_main.patch
+++ b/builder-image-apps/quarkus-vertx/quarkus_main.patch
@@ -1,0 +1,13 @@
+diff --git a/builder-image-apps/quarkus-vertx/pom.xml b/builder-image-apps/quarkus-vertx/pom.xml
+index 4bbe1d3..7d47c9b 100644
+--- a/builder-image-apps/quarkus-vertx/pom.xml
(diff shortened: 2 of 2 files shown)


testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
```diff
@@ -255,6 +255,9 @@ public void debugSymbolsQuarkusContainer(TestInfo testInfo) throws IOException,
             if (QUARKUS_VERSION.startsWith("2.4")) {
                 runCommand(getRunCommand("git", "apply", "quarkus_2.4.x.patch"),
                         Path.of(BASE_DIR, Apps.DEBUG_QUARKUS_BUILDER_IMAGE_VERTX.dir).toFile());
+            } else if (QUARKUS_VERSION.equals("999-SNAPSHOT")) {
```
jerboaa, line 258: How about `... if (QUARKUS_VERSION.contains("SNAPSHOT")) { ...` ? Do we want to hard-code `999`?

zakkak: No opinion on my side, I just went that way since that's the default version used by Quarkus `main` for snapshot builds.

jerboaa: Then just looking for `SNAPSHOT` should be good enough, I think. It's more future proof.

jerboaa (approved):

zakkak:
@Karm kind reminder to review :)

Karm (approved):

Karm:
@zakkak  Hey, I kinda forgot about this. Sorry.
Don't worry about the contains SNAPSHOT  - I have a sort of QuarkusVersion comparator in the making and that would correctly recognize 999-SNAPSHOT as "bigger" than 2.5 and 2.5 as bigger than 2.3 etc.

What is probably wrong is the regexp sanity checking gdb startup again :-( 

```
 Error:  Failures: 
Error:    DebugSymbolsTest.debugSymbolsQuarkus:201 GDB session did not start well. Check the names, paths... Content was: GNU gdb (Ubuntu 10.2-0ubuntu1~20.04~1) 10.2
Copyright (C) 2021 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./target/quarkus-runner...
Reading symbols from /home/runner/work/mandrel-integration-tests/mandrel-integration-tests/ts/apps/quarkus-full-microprofile/target/quarkus-runner.debug...
 ==> expected: <true> but was: <false>
```

PR #70, opened 2021-12-01
CI: Q 2.5.0.Final version bump

zakkak:


.github/workflows/builder_image_tests.yml
```diff
@@ -33,8 +33,10 @@ jobs:
         exclude:
           - quarkus-version: '2.2.3.Final'
             mandrel-builder-image: '20.3-java11'
-          - quarkus-version: '2.4.0.Final'
+          - quarkus-version: '2.5.0.Final'
             mandrel-builder-image: '20.3-java11'
+          - quarkus-version: '2.5.0.Final'
+            mandrel-builder-image: '21.2-java11'
```
jerboaa, line 39: Do we know `2.5.0` doesn't work with the `21.2` image? I assume we only want tests with `21.3`, but still not sure why we exclude 21.2 now.

jerboaa (commented):
Looks OK. What's the reason for the 21.2 exclusion. That's a new change.

.github/workflows/builder_image_tests.yml
zakkak: We know that 2.5.0 no longer supports 21.2, so there is no point in testing it since we won't fix any issues even if they appear (AFAIK it should still work without issues though).

jerboaa: OK

Karm (approved):

Karm (changes requested):
https://github.com/Karm/mandrel-integration-tests/blob/master/testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java#L255

Needs a change. The Quarkus version comparator is still not in PR. I need to get back to that. Sorry for inconvenience.

zakkak:
> https://github.com/Karm/mandrel-integration-tests/blob/master/testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java#L255
> 
> Needs a change. The Quarkus version comparator is still not in PR. I need to get back to that. Sorry for inconvenience.

Fixed in https://github.com/Karm/mandrel-integration-tests/pull/70/commits/2c67a22862ab34e484036ea6c036a793699a7e83 thanks for the info, it saved me quite some time from trying to figure out why it didn't work.

Karm (approved):

PR #71, opened 2021-12-01
Dump processes' output to standard output in addition to the log file

zakkak:
Pros:

1. Makes it easier to detect issues on CI, without having to download the build logs, extract, and look them up.
2. Makes it easier to see that there is progress and the build is not stuck.

Cons:

1. Makes the output a bit noisy.

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
@@ -529,10 +529,11 @@ public class Commands {
                 if (!log.exists()) {
                     Files.createFile(log.toPath());
                 }
-                Files.write(log.toPath(),
-                        ("Command: " + String.join(" ", command) + "\n").getBytes(), StandardOpenOption.APPEND);
-                pb.redirectOutput(ProcessBuilder.Redirect.appendTo(log));
+                String command = "Command: " + String.join(" ", this.command) + "\n";
+                System.out.println(command);
+                Files.write(log.toPath(), command.getBytes(StandardCharsets.UTF_8), StandardOpenOption.APPEND);
                 p = pb.start();
+                dumpAndLogProcessOutput(log, p, timeoutMinutes);
             } catch (IOException e) {
                 e.printStackTrace();
             }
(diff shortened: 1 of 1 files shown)


testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
```diff
@@ -216,6 +216,17 @@ public static void waitForExecutable(List<String> command, File directory) {
         }
     }
 
+    private static void dumpAndLogProcessOutput(File logFile, Process pA) throws IOException {
+        InputStream output = pA.getInputStream();
+        BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(output));
```
jerboaa, line 221: try-with-resources?

testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
```diff
@@ -529,10 +540,11 @@ public void run() {
                 if (!log.exists()) {
                     Files.createFile(log.toPath());
                 }
-                Files.write(log.toPath(),
-                        ("Command: " + String.join(" ", command) + "\n").getBytes(), StandardOpenOption.APPEND);
-                pb.redirectOutput(ProcessBuilder.Redirect.appendTo(log));
+                String command = "Command: " + String.join(" ", this.command) + "\n";
+                System.out.println(command);
+                Files.write(log.toPath(), command.getBytes(), StandardOpenOption.APPEND);
```
jerboaa, line 545: `getBytes(StandardCharsets.UTF-8)`?

jerboaa (approved):
Looks fine a few nits.

testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
zakkak: Good point, thanks!

zakkak:
Converting to draft. It looks like this PR is causing CI to run indefinitely.
:eyes:

zakkak:
The issue causing the stall has been fixed:

The issue was that the new dumpAndLogProcess is blocking, so it can't be used to start a web-server in the background and then start a client to query it. To avoid complicating things, I just don't use dumpAndLogProcess in `runCommand`, in those cases we just redirecto the output like we did before.

Another issue waiting to pop up was that dumpAndLogProcess didn't have a timeout, so if the underlying process was stuck the whole test would get stuck. This is now fixed by starting a single thread executor service for the dumping and logging with a timeout set.

@Karm this is now ready for you to re-review :)

zakkak:
ping @Karm

Karm:
THX @zakkak , I will take a look. I haven't acked it before as I was afraid of what the change does on Jenkins.  Gimme a day.

Karm (approved):
@zakkak Thank you :pray:    It didn't break anything on Jenkins and it seems logged in chronological order it seems.  Merging.

PR #73, opened 2021-12-21
Improve cross-version support with Comparable QuarkusVersion class

zakkak:


Karm (approved):
@zakkak Thank you :pray:  I tested it locally and it works for me nicely.   

My only ask is to keep Mandrel 21.2.0.2 x Quarkus 2.2.3.Final in the GH CI matrix. The point of this workflow is not to test Mandrel or Quarkus but to keep the TS itself compatible with noteworthy M and Q versions...

THX  :fireworks:

zakkak:
I reverted the 2.2 bump and I will merge this once the CI is green again. Thanks Karm!

PR #79, opened 2022-03-22
Add Records annotations test

jerboaa:
Integration test verifying record annotations work. Looking at how GHA respond to this.

PR #84, opened 2022-03-25
[WIP] GH maven settings, Mandrel version checks

Karm:


Karm:
Hmm. It looks nice, but shouldn't it fail with Mandrel  Java 17 - 22.0.0.2-Final ...?    Needs another look.

Karm:
I have no idea :-D It fails locally for me:
 - fails on gdb https://github.com/graalvm/mandrel/issues/355
 - fails on #79 
 
 GH Actions here don't  run those. I am getting bewildered with which source code, from which branch, it actually executes.

Karm:
I'll merge this and restart actions manually on master.

PR #97, opened 2022-06-10
Adds test for spaces in GRAALVM_HOME path

Karm:
Tests https://github.com/graalvm/mandrel/issues/386

Tested both on Linux and Windows.


Karm:
GDB on Ubuntu 20, unrelated to the PR.
Screenshot_20220610-170203

PR #99, opened 2022-06-21
Whitelisting podman warnings

Karm:


Karm:
> Mandrel Locally / ubuntu-20.04 - 11 - 20.3.3.0-Final - 1.11.7.Final 

gdb reg exp no longer compatible with that.

PR #100, opened 2022-07-01
native-image executable file searched for static libs linked

Karm:
The report makes much more sense now:

```
[ERROR] imageioAWTTest{TestInfo}  Time elapsed: 0.121 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 
A different set of static libraries was expected. 
Expected: [libawt.a, libawt_headless.a, libfdlibm.a, libfontmanager.a, libjava.a, libjavajpeg.a, libjvm.a, liblcms.a, liblibchelper.a, libmanagement_ext.a, libnet.a, libnio.a, libzip.a]
Actual:   [libawt.a, libawt_headless.a, libfdlibm.a, libfontmanager.a, libjava.a, libjavajpeg.a, libjvm.a, liblcms.a, liblibchelper.a, libnet.a, libnio.a, libzip.a] ==> expected: <true> but was: <false>
        at org.graalvm.tests.integration.AppReproducersTest.imageioAWT(AppReproducersTest.java:485)
        at org.graalvm.tests.integration.AppReproducersTest.imageioAWTTest(AppReproducersTest.java:359)

```

Let's see if the conditions hold across various Mandrel versions.

Tested locally with:
 * native-image 22.2.0-deve328151c532 Mandrel Distribution (Java Version 17.0.2-internal+0-adhoc.tester.jdk17) :heavy_check_mark: 
 * native-image 22.2.0-deve328151c532 Mandrel Distribution (Java Version 11.0.17-internal+0-adhoc.tester.jdk11) :heavy_check_mark:

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
@@ -52,16 +52,15 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
-import static java.nio.charset.StandardCharsets.US_ASCII;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.graalvm.tests.integration.utils.Commands.builderRoutine;
 import static org.graalvm.tests.integration.utils.Commands.cleanTarget;
 import static org.graalvm.tests.integration.utils.Commands.cleanup;
 import static org.graalvm.tests.integration.utils.Commands.getBaseDir;
 import static org.graalvm.tests.integration.utils.Commands.getRunCommand;
+import static org.graalvm.tests.integration.utils.Commands.listStaticLibs;
 import static org.graalvm.tests.integration.utils.Commands.processStopper;
 import static org.graalvm.tests.integration.utils.Commands.runCommand;
-import static org.graalvm.tests.integration.utils.Commands.searchBinaryFile;
 import static org.graalvm.tests.integration.utils.Commands.searchLogLines;
 import static org.graalvm.tests.integration.utils.Logs.getLogsDir;
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -367,13 +366,11 @@ public class AppReproducersTest {
         LOGGER.info("Testing app: " + app.toString());
 
         final Map<String, String> controlData = new HashMap<>(12);
-
         Process process = null;
         File processLog = null;
         final StringBuilder report = new StringBuilder();
-        final File appDir = new File(BASE_DIR + File.separator + app.dir);
-        final File metaINF = new File(appDir,
-                "src" + File.separator + "main" + File.separator + "resources" + File.separator + "META-INF" + File.separator + "native-image");
+        final File appDir = Path.of(BASE_DIR, app.dir).toFile();
+        final File metaINF = Path.of(BASE_DIR, app.dir, "src", "main", "resources", "META-INF", "native-image").toFile();
         final String cn = testInfo.getTestClass().get().getCanonicalName();
         final String mn = testInfo.getTestMethod().get().getName();
         try {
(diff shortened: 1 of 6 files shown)


testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
```diff
@@ -467,20 +464,26 @@ public void imageioAWT(TestInfo testInfo, Apps app) throws IOException, Interrup
...
+             // libmanagement_ext.a removed again: https://github.com/oracle/graal/pull/4383
+                expected = Set.of("libawt.a", "libawt_headless.a", "libfdlibm.a", "libfontmanager.a",                  "libjava.a", "libjavajpeg.a", "libjvm.a", "liblcms.a", "liblibchelper.a",                         "libnet.a", "libnio.a", "libzip.a");
+            } else if (UsedVersion.jdkFeature(inContainer) > 11 || (UsedVersion.jdkFeature(inContainer) == 11 && UsedVersion.jdkUpdate(inContainer) > 12)) {
                 // Harfbuzz removed: https://github.com/graalvm/mandrel/issues/286
-                toFind = "libnet.a|libjavajpeg.a|libnio.a|liblibchelper.a|libjava.a|liblcms.a|libfontmanager.a|libawt_headless.a|libawt.a|libfdlibm.a|libzip.a|libjvm.a";
+                expected = Set.of("libawt.a", "libawt_headless.a", "libfdlibm.a", "libfontmanager.a",                  "libjava.a", "libjavajpeg.a", "libjvm.a", "liblcms.a", "liblibchelper.a",                         "libnet.a", "libnio.a", "libzip.a");
```
zakkak, line 476: Nitpick, since the order is not important in the Set why don't you append the additional libraries, to avoid the need for the whitespace (which a formatter might remove)

testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
```diff
@@ -467,20 +464,26 @@ public void imageioAWT(TestInfo testInfo, Apps app) throws IOException, Interrup
             // Test static libs in the executable
             final File executable = new File(appDir.getAbsolutePath() + File.separator + "target", "imageio");
             //TODO: This might be too fragile... e.g. order shouldn't matter.
```
zakkak, line 466:
```suggestion
```
This is no longer valid, right?

testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
```diff
@@ -578,26 +581,73 @@ private static void dumpAndLogProcessOutput(File logFile, Process pA, long timeo
...
         }
     }
 
-    public static boolean contains(byte[] one, byte[] theOther) {
+    /**
+     * @param one
+     * @param theOther
+     * @return -1 if one doesn't contain theOther, start index otherwise
+     */
+    public static int contains(byte[] one, byte[] theOther) {
```
zakkak, line 648:
```suggestion
    public static int indexOf(byte[] one, byte[] theOther) {
```

zakkak (changes requested):
LGTM other than some really minor things

jerboaa (approved):
OK, with the suggestions @zakkak pointed out.

testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
Karm: @zakkak I am not sure I understand. I see it as a sort of spreadsheet, immediately seeing which ones are missing.  Do you mean to arrange it in columns as `expected.add()`?

testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
Karm: Done

testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
Karm: @zakkak The whole point of this PR is to make it more understandable, so as one immediately sees what's wrong with the list, so this is actually not a nitpick but an important thing :-D

zakkak: > I see it as a sort of spreadsheet, immediately seeing which ones are missing. 

This only happens in the source code though, right? When printing the Set:
1. The order is not guaranteed
2. the whitespace is not printed

I was thinking about something like:
```
expected = Set.of("libawt.a", "libawt_headless.a", "libfdlibm.a", "libfontmanager.a", "libjava.a", "libjavajpeg.a", "libjvm.a", "liblcms.a", "liblibchelper.a", "libnet.a", "libnio.a", "libzip.a");
```
as the core libraries, and then:
```
expected = Set.of("libawt.a", "libawt_headless.a", "libfdlibm.a", "libfontmanager.a", "libjava.a", "libjavajpeg.a", "libjvm.a", "liblcms.a", "liblibchelper.a", "libnet.a", "libnio.a", "libzip.a", "libharfbuzz.a");
```

or

```
expected = Set.of("libawt.a", "libawt_headless.a", "libfdlibm.a", "libfontmanager.a", "libjava.a", "libjavajpeg.a", "libjvm.a", "liblcms.a", "liblibchelper.a", "libnet.a", "libnio.a", "libzip.a", "libmanagement_ext.a");
```

etc.

>  @zakkak The whole point of this PR is to make it more understandable, so as one immediately sees what's wrong with the list, so this is actually not a nitpick but an important thing :-D

Agreed, wouldn't keeping the "core libraries" as a prefix and then appending the extra libraries for "special" cases have the same result?

zakkak (approved):

Issue #107, opened 2022-08-24
JFRTest fails with graal master: AllowVMInspection deprecated warning


jerboaa:
Discovered here:
https://github.com/graalvm/mandrel/runs/7987554026?check_suite_focus=true

Fails with:

```
Error:  Failures: 
Error:    JFRTest.jfrOptionsSmokeTest:157->jfrOptionsSmoke:288 build-and-run.log log should not contain error or warning lines that are not whitelisted. See /home/runner/work/mandrel/mandrel/mandrel-integration-tests/testsuite/target/archived-logs/org.graalvm.tests.integration.JFRTest/jfrOptionsSmokeTest/build-and-run.log and check these offending lines: 
Warning: Option 'AllowVMInspection' is deprecated and might be removed from future versions: Please use --enable-monitoring ==> expected: <true> but was: <false>
Error:    JFRTest.jfrSmokeTest:83->jfrSmoke:135 build-and-run.log log should not contain error or warning lines that are not whitelisted. See /home/runner/work/mandrel/mandrel/mandrel-integration-tests/testsuite/target/archived-logs/org.graalvm.tests.integration.JFRTest/jfrSmokeTest/build-and-run.log and check these offending lines: 
Warning: Option 'AllowVMInspection' is deprecated and might be removed from future versions: Please use --enable-monitoring ==> expected: <true> but was: <false>
```

we should be using `--enable-monitoring` API option in 22.3+

jerboaa:
@Karm Could you please look at this when you get a chance. Thanks!

jerboaa:
I'll try to get this one fixed.

PR #111, opened 2022-09-22
Hot-replace VM inspection argument

jerboaa:
Depending on the graal vm version in use, use the appropriate native-image option to allow for VM inspection (monitoring).

Closes: #107


testsuite/src/it/java/org/graalvm/tests/integration/utils/versions/GraalVersionProperty.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Sets the property specified by {@code name} (default 'graal.vm.version')
+ * to the currently in-use Graal version for the annotated test run
```
zakkak, line 32: @jerboaa do I understand correctly that this implies that `$JAVA_HOME` is the same as `$GRAALVM_HOME`? Will this still work if we use OpenJDK to run JUNIT and only use GraalVM for the `native-image` part?

jerboaa: I'm not sure, but don't think this is the implication. It seems to assume `native-image` is on the `PATH`, though. It uses [UsedVersion]( https://github.com/Karm/mandrel-integration-tests/blob/master/testsuite/src/it/java/org/graalvm/tests/integration/utils/versions/UsedVersion.java#L36) which is being used elsewhere to determine the GraalVM version (e.g. for the `IfMandrelVersion` annotation). See here:
https://github.com/Karm/mandrel-integration-tests/pull/111/files#diff-e42933e425be03ad47f5f5ede23cae81457941944ed68b35a612ceb10d45c314R38

Does that answer your question?

testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
```diff
@@ -109,6 +127,23 @@ public static String getProperty(String[] alternatives, String defaultValue) {
...
+            String[] tokens = version.split("\\.");
+            if (tokens.length < 2) {
+                return false;
+            }
+            int major = Integer.parseInt(tokens[0]);
+            int minor = Integer.parseInt(tokens[1]);
+            return major >= 22 && minor > 2;
+        } catch (NumberFormatException e) {
+            return false; // assume default of 21
+        }
```
zakkak, line 144: I think you can simplify this with:

```suggestion
        return Version.parse(version).compareTo(22, 3) >= 0;
```

zakkak (approved):
LGTM. Not a blocker, but please consider my suggestion for simplifying the version comparison.

testsuite/src/it/java/org/graalvm/tests/integration/utils/versions/GraalVersionProperty.java
zakkak: Yes, thanks @jerboaa.

testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
jerboaa: Thanks, the latest version uses that now.

Issue #114, opened 2022-09-26
imageioAWT integration test fails with 22.3-dev


jerboaa:
Fails with:
```
[ERROR] Tests run: 11, Failures: 0, Errors: 1, Skipped: 4, Time elapsed: 448.609 s <<< FAILURE! - in org.graalvm.tests.integration.AppReproducersTest
[ERROR] imageioAWTTest{TestInfo}  Time elapsed: 89.856 s  <<< ERROR!
java.lang.UnsupportedOperationException
	at org.graalvm.tests.integration.AppReproducersTest.imageioAWT(AppReproducersTest.java:526)
	at org.graalvm.tests.integration.AppReproducersTest.imageioAWTTest(AppReproducersTest.java:417)
```

Also seen in CI here:
https://github.com/graalvm/mandrel/actions/runs/3127680442/jobs/5074987920#step:10:18615

jerboaa:
Sigh, caused by #112 which adds an entry to the unmodifiable Set created by `Set.of`.

jerboaa:
Interestingly we'd have seen this for the `libharfbuzz.a` case too. I guess we just didn't hit that JDK case for 11.

PR #115, opened 2022-09-26
Fix adding to unmodifiable Set

jerboaa:
Closes: #114


jerboaa:
I don't think test failures are related to this patch.

zakkak (approved):
Thanks @jerboaa

PR #121, opened 2022-10-05
Fail instead of throwing NPE when Process is null

Karm:
Small UX improvement: If the executable fails to start, e.g. native image not built, it throws an NPE. JUnit failure is better in that case to understand for an uninitiated reader.


zakkak (approved):
LGTM, thanks @Karm

Issue #110, opened 2022-09-22
DebugSymbolsTest fails with Mandrel 22.3.0-dev


jerboaa:
Fails with:

```
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 340.424 s <<< FAILURE! - in org.graalvm.tests.integration.DebugSymbolsTest
[ERROR] debugSymbolsSmokeGDB{TestInfo}  Time elapsed: 72.936 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 
There were errors in the GDB session. Note that commands in the session might depend on each other. Errors: 
Command 'info functions .*smoke.*' did not match the expected pattern '.*File debug_symbols_smoke/ClassA.java:.*java.lang.String \*debug_symbols_smoke.ClassA::toString\(void\).*File debug_symbols_smoke/Main.java:.*void debug_symbols_smoke.Main\$\$Lambda\$.*::accept\(java.lang.Object \*\).*void debug_symbols_smoke.Main::lambda\$thisIsTheEnd\$0\(java.io.ByteArrayOutputStream \*, debug_symbols_smoke.ClassA \*\).*void debug_symbols_smoke.Main::main\(java.lang.String\[\] \*\).*void debug_symbols_smoke.Main::thisIsTheEnd\(java.util.List \*\).*'.
Output was:
(gdb) All functions matching regular expression ".*smoke.*":

File debug_symbols_smoke/ClassA.java:
37:	java.lang.String *debug_symbols_smoke.ClassA::toString(debug_symbols_smoke.ClassA *);

File debug_symbols_smoke/Main.java:

    	void debug_symbols_smoke.Main$$Lambda$768/0x00000007c08fcc18::accept(debug_symbols_smoke.Main$$Lambda$768/0x00000007c08fcc18 *, java.lang.Object *);

    50:	void debug_symbols_smoke.Main::lambda$thisIsTheEnd$0(java.io.ByteArrayOutputStream *, debug_symbols_smoke.ClassA *);
64:	void debug_symbols_smoke.Main::main(java.lang.String[] *);
49:	void debug_symbols_smoke.Main::thisIsTheEnd(java.util.List *);
 ==> expected: <true> but was: <false>
	at org.graalvm.tests.integration.DebugSymbolsTest.carryOutGDBSession(DebugSymbolsTest.java:385)
[...]

PR #122, opened 2022-10-10
Fix debug smoke test after improved mangle support

jerboaa:
https://github.com/oracle/graal/pull/5172 fixed this upstream issue https://github.com/oracle/graal/issues/5064. However, the test still asserted wrong data. Now stripped spaces in function parameters and not including a specific 'void' string for void functions.

Closes: #110


zakkak (approved):
Thanks @jerboaa!

PR #127, opened 2023-01-04
Adds perf smoke tests, report and fixes incompatibilities

Karm:
Adds a tiny dead simple Q app that just serializes/deserializes a class from JSON, uses a chunk of static String array to do some basic strings ops and hashes string with SHA-256.

Test class uses it to make it serialize and deserialize ~200MB of JSON, it measures the runtime perf characteristics like cpu cycles, page misses, GC events etc. using `perf` tool and GC logging.

It spits out results such as:
```
[
  {
    "arch": "amd64",
    "branchMisses": 365341738,
    "branches": 23826053769,
    "contextSwitches": 63309,
    "cpuMigrations": 531,
    "cycles": 75202164737,
    "executableSizeKb": "-1",
    "file": "java -Xlog:gc -XX:+UseSerialGC -Xmx2560m -jar target/quarkus-app/quarkus-run.jar",
    "fullGCevents": "-1",
    "incrementalGCevents": "-1",
    "instructions": 137792790971,
    "jdkVersion": "17.0.6",
    "mandrelVersion": "23.0.0-devb6e1c8d0844",
    "maxHeapSizeMB": 2560,
    "os": "Linux",
    "pageFaults": 474282,
    "parseOnce": "null",
    "quarkusVersion": "2.13.6.Final",
    "ramAvailableMB": 7295,
    "rssKb": 2163884,
    "runnerDescription": "",
    "secondsTimeElapsed": "17.036505968",
    "taskClock": "24932.19",
    "testApp": "https://github.com/Karm/mandrel-integration-tests/apps/quarkus-json/",
    "timeSpentInGCs": "7.2116430000000005"
  },
  {
    "arch": "amd64",
    "branchMisses": 340702165,
    "branches": 69657566354,
    "contextSwitches": 74753,
    "cpuMigrations": 428,
    "cycles": 129039893879,
    "executableSizeKb": 47292,
    "file": "./target/quarkus-json-runner -XX:+PrintGC",
    "fullGCevents": 23,
    "incrementalGCevents": 61,
    "instructions": 334111941045,
    "jdkVersion": "17.0.6",
    "mandrelVersion": "23.0.0-devb6e1c8d0844",
    "maxHeapSizeMB": 2560,
    "os": "Linux",
    "pageFaults": 3283541,
    "quarkusVersion": "2.13.6.Final",
    "ramAvailableMB": 7054,
    "rssKb": 1903552,
    "runnerDescription": "",
    "secondsTimeElapsed": "42.657960072",
    "taskClock": "43140.99",
    "testApp": "https://github.com/Karm/mandrel-integration-tests/apps/quarkus-json/",
    "timeSpentInGCs": "25.300015799999997"
  }
]
```

The result ain't uploaded anywhere ATM, I will open a PR in Collector for that presently...

Then it would pop up on a dashboard similar to this demo: https://karms.biz/tmp/collector-gui/

Collector usage:

```
 mvn  clean verify -Ptestsuite -DexcludeTags=all -DincludeTags=perfcheck -Dquarkus.version=2.13.6.Final  -Dperf.app.report=true -Dperf.app.endpoint=http://localhost:8080 -Dperf.app.secret.token=token
```


apps/quarkus-json/src/main/java/org/acme/Record.java
```diff
@@ -0,0 +1,26 @@
...
+    public String s7;
+    public String s8;
+    public String s9;
+    public int i0;
+    public int i1;
+    public int i2;
+    public int i3;
+    public int i4;
+    public int i5;
+    public int i6;
```
zakkak, line 23: Would it make sense to include some more types to increase coverage?

testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
```diff
@@ -50,6 +54,40 @@ public enum BuildAndRunCmds {
             new String[]{"mvn", "dependency:sources", "-Dquarkus.version=" + QUARKUS_VERSION.getVersionString()},
             new String[]{IS_THIS_WINDOWS ? "target\\quarkus-runner.exe" : "./target/quarkus-runner"}
     }),
+    QUARKUS_JSON_PERF_PARSEONCE(new String[][]{
```
zakkak, line 57: Maybe we should call this PARSETWICE as parseonce is the default :)

testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
```diff
@@ -50,6 +54,40 @@ public enum BuildAndRunCmds {
...
+                            "-H:-ParseOnce," +
+                            "-H:BuildOutputJSONFile=quarkus-json_minus-ParseOnce.json",
+                    "-Dfinal.name=quarkus-json_-ParseOnce"},
+            new String[]{"mvn", "package", "-Pnative", "-Dquarkus.version=" + QUARKUS_VERSION.getVersionString(),
+                    "-Dquarkus.native.additional-build-args=" +
+                            ((UsedVersion.getVersion(false).compareTo(Version.create(23, 0, 0)) >= 0) ? "-H:+AllowDeprecatedBuilderClassesOnImageClasspath," : "") +
+                            "-R:MaxHeapSize=" + MX_HEAP_MB + "m," +
+                            "-H:+ParseOnce," +
+                            "-H:BuildOutputJSONFile=quarkus-json_plus-ParseOnce.json",
+                    "-Dfinal.name=quarkus-json_+ParseOnce"},
```
zakkak, line 72: I suggest avoiding the duplication here since only `ParseOnce` changes, to make it easier to keep the rest the same.

testsuite/src/test/resources/parse-serial-gc-build-and-run.log
```diff
@@ -0,0 +1,1066 @@
...
+Warning: Class-path entry file:///home/karm/workspaceRH/mandrel-integration-tests-dev/apps/quarkus-json/target/quarkus-json_-ParseOnce-native-image-source-jar/lib/org.graalvm.sdk.graal-sdk-22.3.0.jar contains class com.oracle.svm.core.annotate.TargetElement. This class is part of the image builder itself (in jrt:/org.graalvm.sdk) and must not be passed via -cp.
```
zakkak, line 40: What does it take to get rid of these warnings ?

apps/quarkus-json/src/main/java/org/acme/Record.java
Karm: Not sure, I will take a look what the serializer/deserializer does with e.g. BigInt etc. Perhaps calendar dates...some more complex types. The core idea is string processing though. Maybe we can have more small apps?

testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
Karm: I will clean it up.

testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
```diff
@@ -50,6 +54,40 @@ public enum BuildAndRunCmds {
...
+            //-XX:+UseShenandoahGC
+            //-XX:+UseSerialGC
+            //-XX:+UseG1GC
+            new String[]{"perf", "stat", "--delay", "2000", "java", "-Xlog:gc", "-XX:+UseSerialGC", "-Xmx" + MX_HEAP_MB + "m", "-jar", "target/quarkus-app/quarkus-run.jar"},
```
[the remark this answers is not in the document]
Karm: While the comparison is not apples to apples, using serial gc is at least an attempt to make it fruit to fruit :)  
I understand that the SerialGC in Svm is a completely different implementation than SerialGC in HotSpot, yet it levels the filed a bit.
The processing of the test data is much faster with HotSpot than with Native and if you use G1 or Shenandoah the gap widens even more as Native is stuck with SerialGC...

testsuite/src/test/resources/parse-serial-gc-build-and-run.log
Karm: I think we need to address that on the Quarkus side. It is what you get when running Q2.13 with the GraalVM 23 dev.

IMHO:  
 * https://github.com/oracle/graal/commit/90c1f568d629579abee670d369faf91e0847df6c
 * https://github.com/oracle/graal/blob/b6e1c8d0844c475c97ad7ef52796b192b59ef2fc/substratevm/CHANGELOG.md?plain=1#L6

testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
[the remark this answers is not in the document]
zakkak: What's the purpose of these checks? Is it to see if GC is becoming slower in native? And do I understand correctly that if that's the case, it will be easier to observe this if we compare with a GC algorithm closer to what we are measuring?

testsuite/src/test/resources/parse-serial-gc-build-and-run.log
zakkak: > I think we need to address that on the Quarkus side.

I have actually fixed this already... https://github.com/quarkusio/quarkus/pull/28650

I will need to start doing some memory exercises...

testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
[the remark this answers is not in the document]
Karm: The purpose is to take HotSpot perf as a kind of a baseline and measure how native SVM perf differs.
To make sure the gap is not widening, preferably for us to be able to even pursue closing that gap.

By perf I mean a matrix of indicators we could optimize for, understanding the tradeoffs.
 - overall time spent processing a given workload is the most obvious one
 - CPU cycles
 - instructions
 - RSS
 - time spent doing GC

So while the SerialGC in SVM is a copying one and the SerialGC in HotSpot is a mark-sweep-compact one, it still makes sense to me to compare the overall time spent doing GC.

On the other hand, the intermediate GC events, partial GC, are not recorded for HotSpot as there is nothing to compare those events to in native run...

This whole thing started for me as a curious poking in `perf` and `callgrind` to observe differences between setting and not setting `ParseOnce`. So seeking runtime perf differences among native-image executables is another valid angle IMHO.

testsuite/src/test/resources/parse-serial-gc-build-and-run.log
Karm: THX :pray: :+1:

testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
```diff
@@ -68,28 +72,32 @@
...
+    public static final boolean FAIL_ON_PERF_REGRESSION = Boolean.parseBoolean(getProperty("FAIL_ON_PERF_REGRESSION", "true"));
+
     public static final boolean IS_THIS_WINDOWS = System.getProperty("os.name").matches(".*[Ww]indows.*");
     private static final Pattern NUM_PATTERN = Pattern.compile("[ \t]*[0-9]+[ \t]*");
     private static final Pattern ALPHANUMERIC_FIRST = Pattern.compile("([a-z0-9]+).*");
     private static final Pattern CONTAINER_STATS_MEMORY = Pattern.compile("(?:table)?[ \t]*([0-9\\.]+)([a-zA-Z]+).*");
 
-    public static String getProperty(String[] alternatives, String defaultValue) {
+    public static final String GRAALVM_ALLOW_DEPRECATED_BUILDER_SWITCH_TOKEN = "<GRAALVM_ALLOW_DEPRECATED_BUILDER>";
+    public static final String GRAALVM_ALLOW_DEPRECATED_BUILDER_SWITCH = "-H:+AllowDeprecatedBuilderClassesOnImageClasspath";
```
Karm, line 87: @zakkak So I guess I can get rid of this now...?

testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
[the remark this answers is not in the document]
zakkak: ACK, thanks for the clarification

testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
zakkak: Yes, but only for Quarkus >= 2.14. Which makes sense since 2.13 only supports GraalVM/Mandrel 22.3

Karm:
fixes #132 in [7ef2f6e](https://github.com/Karm/mandrel-integration-tests/pull/127/commits/7ef2f6ee52a7290b7a8f57664a077dcf81012d0e) and [991c5c4](https://github.com/Karm/mandrel-integration-tests/pull/127/commits/991c5c40bbe07a13b03b079c99121ae3c85cecd1)

Karm:
[46a3afa](https://github.com/Karm/mandrel-integration-tests/pull/127/commits/46a3afa0d411070eddf1c3da1d6ff74767b876ad) fixes #130

Karm:
[662b77a](https://github.com/Karm/mandrel-integration-tests/pull/127/commits/662b77a68b7ec1a9edd9baa949a8e73d3988ad07) fixes #132

Karm:
[f3588ae](https://github.com/Karm/mandrel-integration-tests/pull/127/commits/f3588ae4b9d5d9eefe3c02cd19aef1d7cb13f96a) adapts to quarkusio/quarkus/issues/30998

Karm:
@zakkak  I do realize it would have been cleaner if I separated the incompatibility fixes, yet this was an ample opportunity to untangle some accumulated incompatibilities at once, without prolonging the pain.

Namely the separation of JVM builds and  Native builds, Jakarta move, Temurin JDK 20 ea version parsing, all enables to run the perf results and let the reporter upload those now...

apps/quarkus-json/src/main/java/org/acme/ExampleResource.java
```diff
@@ -0,0 +1,220 @@
+package org.acme;
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
```
jerboaa, line 6: Why would this not need the `jakarta` import changes similar to what is being done in  `quarkus-3.patch`?

jerboaa (commented):
One question:

apps/quarkus-json/src/main/java/org/acme/ExampleResource.java
Karm: @jerboaa Very good catch.  The app is used only in PerfCheck tests and those are not on by default, so the CI here didn't catch it.  ...and my Jenkins job doesn't perf tests Q 3.x yet.
I will add PerfCheck tests to the CI in this repo so as everything is covered.

Lemme patch it.

jerboaa: Thanks!

jerboaa (approved):
Seems OK. GHA seems to have failures though. Please check before you merge.

.github/workflows/local_tests.yml
```diff
@@ -48,13 +48,23 @@ jobs:
...
             java-version: '11'
 
-          - quarkus-version: '2.16.0.Final'
+          - quarkus-version: '2.16.1.Final'
             mandrel-version: '21.3.5.1-Final'
             java-version: '11'
-          - quarkus-version: '2.16.0.Final'
+          - quarkus-version: '2.16.1.Final'
             mandrel-version: '21.3.5.1-Final'
             java-version: '17'
```
zakkak, line 56: I think that you can just do:

```
exclude:
          - quarkus-version: '2.16.1.Final'
            mandrel-version: '21.3.5.1-Final'
```

and it will exclude both Java 11 and 17.

See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixexclude

apps/quarkus-full-microprofile/quarkus_3.x.patch
```diff
@@ -0,0 +1,215 @@
...
++import jakarta.enterprise.context.ApplicationScoped;
++import jakarta.ws.rs.GET;
++import jakarta.ws.rs.Path;
++import jakarta.ws.rs.WebApplicationException;
++import jakarta.ws.rs.client.ClientBuilder;
++import jakarta.ws.rs.client.WebTarget;
++import jakarta.ws.rs.core.Response;
+ import java.io.BufferedReader;
+ import java.io.IOException;
+ import java.io.InputStreamReader;
```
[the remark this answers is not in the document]
Karm: I'd stick to patch because we already have a mechanism that also reverts the changes with `-R` in finally block.

zakkak (approved):
LGTM other than the CI failures. Thanks @Karm

Karm:
Looks like `perf` can work. The system was just too restrictive. It is failing on timeouts now. I will lift the timeout and try again...

Karm:
@zakkak @jerboaa  Finally all green 🟩✔️

PR #128, opened 2023-01-25
`--enable-monitoring` without arguments deprecated

zakkak:


testsuite/src/it/java/org/graalvm/tests/integration/JFRTest.java
```diff
@@ -75,6 +75,7 @@ public class JFRTest {
     public static final String BASE_DIR = getBaseDir();
 
     public enum JFROption {
+        MONITOR_23("--enable-monitoring=all"),
         MONITOR_22("--enable-monitoring"),
```
jerboaa, line 79: AFAIK, 22.3 supports `--enable-monitoring=all` as well. Should we just use it for 22.3 onwards instead of doing another conditional for `MONITOR_23`?

```
$ native-image --version
native-image 22.3.1 Mandrel Distribution (Java Version 17.0.6-internal+0-adhoc.sgehwolf.jdk17u)
$ native-image --help
[...]
    --enable-monitoring   enable monitoring features that allow the VM to be inspected at
                          run time. Comma-separated list can contain 'heapdump', 'jfr',
                          'jvmstat', or 'all' (defaults to 'all' if no argument is
                          provided). For example: `--enable-monitoring=heapdump,jvmstat`.
[...]
```

jerboaa (changes requested):
We should consider doing `--enable-monitoring=all` for `>= 22.3`. Thoughts?

jerboaa:
Or better yet, use `--enable-monitoring=jfr` as we are using this test for JFR only?

jerboaa (approved):
LGTM

PR #136, opened 2023-02-14
Q 999-SNAPSHOT needs GIT SHA in report

### LINKED ISSUE #135: PerfCheck: Distinguish between various 999-SNAPSHOT Q versions
The current [comparator](https://github.com/graalvm/mandrel-packaging/blob/master/jenkins/jobs/tests/mandrel_linux_integration_tests_perf_comparator.groovy) Jenkins job takes two Quarkus git sha commits and runs perf report with both.

The problem is that both get reported as 999-SNAPSHOT:

screenshot

TODO:
 - either add a new column with...some qualifier?
 - or append git sha to the `999-` ?  (sounds better)

Karm:
fixes #135 

It works as expected, I have this in DB now:

screenshot

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/versions/QuarkusVersion.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/versions/QuarkusVersion.java
@@ -19,6 +19,8 @@
  */
 package org.graalvm.tests.integration.utils.versions;
 
+import static org.graalvm.tests.integration.utils.Commands.getProperty;
+
 public class QuarkusVersion implements Comparable<QuarkusVersion> {
 
     private final String version;
@@ -26,6 +28,11 @@ public class QuarkusVersion implements Comparable<QuarkusVersion> {
     private final int minor;
     private final int patch;
     private final boolean snapshot;
+    private final String gitSHA;
+
+    public QuarkusVersion() {
+        this(getProperty("QUARKUS_VERSION", "2.13.7.Final"));
+    }
 
     public QuarkusVersion(String version) {
         this.version = version;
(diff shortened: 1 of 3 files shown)


testsuite/src/it/java/org/graalvm/tests/integration/utils/versions/QuarkusVersion.java
```diff
@@ -34,11 +41,13 @@ public QuarkusVersion(String version) {
             this.major = Integer.parseInt(version.split("-")[0]);
             this.minor = 0;
             this.patch = 0;
+            this.gitSHA = getProperty("QUARKUS_VERSION_GITSHA", "");
```
zakkak, line 44: Where is this being set though? Does it happen in some script outside of this repo?

zakkak (approved):
LGTM, I just have a question about where is `QUARKUS_VERSION_GITSHA` being set.

testsuite/src/it/java/org/graalvm/tests/integration/utils/versions/QuarkusVersion.java
Karm: It's the same as with all other properties, including `QUARKUS_VERSION` or e.g. `FAIL_ON_PERF_REGRESSION`, i.e. outside the repo, in the environment that runs the TS.

`QUARKUS_VERSION_GITSHA` specifically is currently used in a one test job only: [mandrel-linux-integration-tests-perf-comparator](https://ci.modcluster.io/view/Mandrel/job/mandrel-linux-integration-tests-perf-comparator/), [source](https://github.com/graalvm/mandrel-packaging/blob/master/jenkins/jobs/tests/Constants.groovy#L112), [job definition](https://github.com/graalvm/mandrel-packaging/blob/master/jenkins/jobs/tests/mandrel_linux_integration_tests_perf_comparator.groovy).

Issue #148, opened 2023-04-12
UsedVersion.InContainer fails cryptically when image not pulled


jerboaa:
Example failure:

```
Error:  Errors: 
Error:    AppReproducersTest.imageioAWTContainerTest » NoClassDefFound Could not initial...
Error:    JFRTest.jfrOptionsSmokeContainerTest » ExceptionInInitializer
Error:    JFRTest.jfrSmokeContainerTest » NoClassDefFound Could not initialize class org...
```

Seen here: https://github.com/graalvm/mandrel/actions/runs/4642679968/jobs/8283797496#step:10:11195

Reproducible with something like this:

```
$ podman rmi quay.io/quarkus/ubi-quarkus-mandrel-builder-image:22.3-java17
$ export QUARKUS_NATIVE_CONTAINER_RUNTIME="podman"
$ export QUARKUS_NATIVE_BUILDER_IMAGE="quay.io/quarkus/ubi-quarkus-mandrel-builder-image:22.3-java17"
$ mvn clean verify -Dquarkus.version=999-SNAPSHOT -Ptestsuite-builder-image -Dtest=AppReproducersTest#imageioAWTContainerTest
```

Fails then with:

```
2023-04-12 19:01:33.425 INFO  [o.g.t.i.u.v.UsedVersion$MVersion] (runNativeImageVersion) Running command [podman, run, -t, quay.io/quarkus/ubi-quarkus-mandrel-builder-image:22.3-java17, native-image, --version] to determine Mandrel version used.
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 71.368 s <<< FAILURE! - in org.graalvm.tests.integration.AppReproducersTest
[ERROR] imageioAWTContainerTest{TestInfo}  Time elapsed: 0.007 s  <<< ERROR!
java.lang.ExceptionInInitializerError
Caused by: java.lang.NullPointerException

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Errors: 
[ERROR]   AppReproducersTest.imageioAWTContainerTest » ExceptionInInitializer
[INFO] 
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Native image integration TS 1.0.0-SNAPSHOT:
[INFO] 
[INFO] Native image integration TS ........................ SUCCESS [  0.094 s]
[...]

PR #149, opened 2023-04-12
Pull the image before running --version

jerboaa:
This avoids the output getting polluted when `native-image --version` runs and it's output is getting parsed for version info.

Closes: #148


Karm:
THX :pray:

Issue #151, opened 2023-04-20
imageio testfix after dynamic AWT change for 23.0 was incomplete


jerboaa:
After https://github.com/Karm/mandrel-integration-tests/pull/146, we still see imageio test failures in CI:

```
2023-04-12 11:12:41.242 INFO  [o.g.t.i.AppReproducersTest] (imageioAWT) Running...
Error:  Tests run: 12, Failures: 1, Errors: 0, Skipped: 5, Time elapsed: 576.106 s <<< FAILURE! - in org.graalvm.tests.integration.AppReproducersTest
Error:  imageioAWTTest{TestInfo}  Time elapsed: 133.032 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 
There were errors checking the generated image files, see:
mytest.jpg was not generated.
mytest.bmp was not generated.
mytest.gif was not generated.
mytest.svg was not generated.
mytest.png was not generated.
mytest.tiff was not generated.
mytest_Resized_Grace_M._Hopper.png was not generated.
mytest.wbmp was not generated. ==> expected: <true> but was: <false>
	at org.graalvm.tests.integration.AppReproducersTest.imageioAWT(AppReproducersTest.java:594)
	at org.graalvm.tests.integration.AppReproducersTest.imageioAWTTest(AppReproducersTest.java:494)

[INFO] 
[INFO] Results:
[INFO] 
Error:  Failures: 
Error:    AppReproducersTest.imageioAWTTest:494->imageioAWT:594 There were errors checking the generated image files, see:
mytest.jpg was not generated.
mytest.bmp was not generated.
mytest.gif was not generated.
mytest.svg was not generated.
mytest.png was not generated.
mytest.tiff was not generated.
mytest_Resized_Grace_M._Hopper.png was not generated.
mytest.wbmp was not generated. ==> expected: <true> but was: <false>
[INFO] 
Error:  Tests run: 18, Failures: 1, Errors: 0, Skipped: 5
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Native image integration TS 1.0.0-SNAPSHOT:
[INFO] 
[INFO] Native image integration TS ........................ SUCCESS [  0.882 s]
[INFO] testsuite .......................................... FAILURE [29:40 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  29:41 min
[...]

PR #152, opened 2023-04-24
Adds fake JAVA_HOME to appease FontConfig

Karm:
It works both locally:

```
mvn clean verify -Ptestsuite -Dtest=AppReproducersTest#imageioAWTTest
```
and in a container:

```
mvn clean verify -Ptestsuite-builder-image -Dtest=AppReproducersTest#imageioAWTContainerTest -Dquarkus.native.builder-image=quay.io/karmkarm/ubi-quarkus-mandrel-builder-image:23.0-java20
```

How come it works in the container without creating those JAVA_HOME dirs in the container and copying .so files there?
It's because the TS mounts the app dir, parent o the target dir:

```
docker run -u 1000:1000 -t -v /home/karm/workspaceRH/mandrel-integration-tests-Apr2023/apps/imageio:/work:z my-imageio-runner /work/target/imageio -Djava.home=. -Djava.awt.headless=true
```

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
@@ -551,6 +551,12 @@ public class AppReproducersTest {
             controlData.keySet().forEach(f -> new File(appDir, f).delete());
 
             LOGGER.info("Running...");
+
+            // Fontconfig might look for fonts here, see similar thing in Quarkus:
+            // https://github.com/quarkusio/quarkus/blob/main/extensions/awt/runtime/src/main/java/io/quarkus/awt/runtime/JDKSubstitutions.java#L53
+            // Details: https://github.com/Karm/mandrel-integration-tests/issues/151#issuecomment-1516802244
+            Files.createDirectories(Path.of(appDir.toString(), "lib")).toFile().deleteOnExit();
+
             final List<String> cmd = getRunCommand(app.buildAndRunCmds.cmds[app.buildAndRunCmds.cmds.length - 1]);
             process = runCommand(cmd, appDir, processLog, app);
             assertNotNull(process, "The test application failed to run. Check " + getLogsDir(cn, mn) + File.separator + processLog.getName());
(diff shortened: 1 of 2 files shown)


testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
```diff
@@ -551,6 +551,12 @@ public void imageioAWT(TestInfo testInfo, Apps app) throws IOException, Interrup
             controlData.keySet().forEach(f -> new File(appDir, f).delete());
 
             LOGGER.info("Running...");
+
+            // Fontconfig might look for fonts here, see similar thing in Quarkus:
+            // https://github.com/quarkusio/quarkus/blob/main/extensions/awt/runtime/src/main/java/io/quarkus/awt/runtime/JDKSubstitutions.java#L53
+            Files.createDirectories(Path.of(appDir.toString(), "conf", "fonts")).toFile().deleteOnExit();
+            Files.createDirectories(Path.of(appDir.toString(), "lib")).toFile().deleteOnExit();
```
jerboaa, line 558: I don't think we need both. One of the two should be sufficient. See:
https://github.com/Karm/mandrel-integration-tests/issues/151
and
https://github.com/openjdk/jdk17u/blob/908cab4123812b6b206b966a6ce92398cdf42c08/src/java.desktop/share/classes/sun/awt/FontConfiguration.java#L183 and https://github.com/openjdk/jdk17u/blob/908cab4123812b6b206b966a6ce92398cdf42c08/src/java.desktop/share/classes/sun/awt/FontConfiguration.java#L193

I suggest to only create the `lib` directory.

jerboaa (changes requested):
Please link https://github.com/Karm/mandrel-integration-tests/issues/151 to this PR.

testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
```diff
@@ -551,6 +551,11 @@ public void imageioAWT(TestInfo testInfo, Apps app) throws IOException, Interrup
             controlData.keySet().forEach(f -> new File(appDir, f).delete());
 
             LOGGER.info("Running...");
+
+            // Fontconfig might look for fonts here, see similar thing in Quarkus:
+            // https://github.com/quarkusio/quarkus/blob/main/extensions/awt/runtime/src/main/java/io/quarkus/awt/runtime/JDKSubstitutions.java#L53
```
jerboaa, line 556: Perhaps link to #151 as well? It has some more details.

jerboaa (approved):
LGTM.

Issue #154, opened 2023-05-08
DebugSymbolsTest.debugSymbolsQuarkus fails with latest GraalVM master


jerboaa:
debugSymbolsSmokeGDB fails with (both with Java 17 and Java 20):

```
Error:  Tests run: 2, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 724.231 s <<< FAILURE! - in org.graalvm.tests.integration.DebugSymbolsTest
Error:  debugSymbolsQuarkus{TestInfo}  Time elapsed: 543.363 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 
There were errors in the GDB session. Note that commands in the session might depend on each other. Errors: 
Command 'b ConfigTestController.java:33' did not match the expected pattern '.*Breakpoint 1 at .*: file com/example/quarkus/config/ConfigTestController.java, line 33.*'.
Output was:
(gdb) (gdb) (gdb) GNU gdb (Ubuntu 12.1-0ubuntu1~22.04) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word".
 ==> expected: <true> but was: <false>
	at org.graalvm.tests.integration.DebugSymbolsTest.carryOutGDBSession(DebugSymbolsTest.java:394)
	at org.graalvm.tests.integration.DebugSymbolsTest.debugSymbolsQuarkus(DebugSymbolsTest.java:219)

Error:  debugSymbolsSmokeGDB{TestInfo}  Time elapsed: 180.857 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 
There were errors in the GDB session. Note that commands in the session might depend on each other. Errors: 
Command 'break Main.java:71' did not match the expected pattern '.*Breakpoint 2 at .*: file debug_symbols_smoke/Main.java, line 71.*'.
Output was:
[...]

PR #155, opened 2023-05-08
Update flags for DebugSymbolsSmoke test

zakkak:
New defaults (see https://github.com/oracle/graal/pull/6504) coming in 23.1 (and possibly 23.0?)

Closes: #154


testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
```diff
@@ -203,6 +206,12 @@ public enum BuildAndRunCmds {
                     new String[]{"powershell", "-c", "\"Expand-Archive -Path test_data.txt.zip -DestinationPath target -Force\""}
                     :
                     new String[]{"unzip", "test_data.txt.zip", "-d", "target"},
+            UsedVersion.getVersion(false).compareTo(Version.create(23, 1, 0)) >= 0 ?
```
jerboaa, line 209: I don't think we can use `UsedVersion` like this. We'll have to add some tokens a la `JFR_MONITORING_SWITCH_TOKEN` and replace as appropriate for various versions. CI failures I think are also because of this, since we are hard-coding the used version to be non-container.

jerboaa (changes requested):

testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
zakkak: Indeed

testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
```diff
@@ -80,6 +85,20 @@ public class DebugSymbolsTest {
 
     public static final String BASE_DIR = getBaseDir();
 
+    public enum DebugOptions {
+        DEBUG_23_0("<DEBUG_FLAGS_23_0>", "-H:+TrackNodeSourcePosition"),
+        DEBUG_23_1_a("<DEBUG_FLAGS_23_1_a>", "-H:+DebugCodeInfoUseSourceMappings"),
+        DEBUG_23_1_b("<DEBUG_FLAGS_23_1_b>", "-H:+OmitInlinedMethodDebugLineInfo");
```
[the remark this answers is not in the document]
zakkak: Done. I kept the "version suffix"  to easily see in which version they started being required.

jerboaa (approved):

PR #156, opened 2023-05-09
Refactor BuildAndRunCmds to semantically separate build from run commands

zakkak:


zakkak:
@Karm could you please have a look? Would something like this make sense? Shall I rebase?

Karm:
Hello @zakkak, in general I find it a great idea to untangle the semantics of what are essentially **build** commands, being it `mvn` or `native-image` or direct calls to `jar` etc.  from  what are actually **run** commands, e.g. calls to container engine to start something, to run an executable, to run `java` command directly.

Tests currently rely on a fact that these commands have dependency `n+1` on `n` etc. Some tests run `n`, `n+1`, collect something and then run `n+2` and `n+3`. 

If this could be preserved so as things don't blow up, I'd be very happy for a better semantics. Hopefully two categories, building stuff and running stuff? 

TL;DR:  Yes, let's proceed if you have some cycles for this and if it makes your life with the TS easier.

zakkak:
Rebased, let's see what the CI says. I had to split the `monitorFieldOffset` test since it was the only one doing `build -> run -> build -> run`.

Karm (changes requested):
At least in case of JFR smoke tests, there is a wrong flow of commands :(  Off by one?

zakkak:
@Karm PTAL, the CI look happy now. I found and fixed two issues.

Karm (approved):
Thx. CI is happy, it looks sane. works locally. If there is a bug we need to fix the CI to catch it :)

Issue #157, opened 2023-05-10
Revisit how gdb output is checked


zakkak:
GDB output is checked with a timeout of 10 seconds, which seems reasonable:

https://github.com/Karm/mandrel-integration-tests/blob/eb6a37ad682bf3ed9c92362d89cc80b7f38b9912/testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java#L414

The issue is that apparently this results in misleading error messages on timeout, e.g.: 

In https://github.com/graalvm/mandrel/actions/runs/4932214298/jobs/8815389640#step:10:19429 we get

```
Command 'b ConfigTestController.java:33' did not match the expected pattern '.*Breakpoint 1 at .*: file com/example/quarkus/config/ConfigTestController.java, line 33.*'.
Output was:
(gdb) (gdb) (gdb) GNU gdb (Ubuntu 12.1-0ubuntu1~22.04) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word".
```

The output that is shown is from the previous gdb command `show version`.

My understanding is that `waitForBufferToMatch` returns immediately in the case of `show version` since the pattern will match anything (even no output), as a result if `gdb` takes more than 10 seconds to run `b ConfigTestController.java:33` and produces no output we will most probably observe the output from `show version`.

PR #158, opened 2023-05-10
Show version in gdb should return text including `gdb`

zakkak:
Waiting for `.*` results in essentially matching no output as well, which is not intended.

Fixes: #157

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
@@ -272,7 +272,7 @@ public enum GDBSession {
         }
     };
 
-    private static final CP SHOW_VERSION = new CP("show version\n", Pattern.compile(".*", Pattern.DOTALL));
+    private static final CP SHOW_VERSION = new CP("show version\n", Pattern.compile(".*gdb.*", Pattern.DOTALL));
 
     public abstract CP[] get(boolean inContainer);
 }


testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
```diff
@@ -272,7 +272,7 @@ public CP[] get(boolean inContainer) {
         }
     };
 
-    private static final CP SHOW_VERSION = new CP("show version\n", Pattern.compile(".*", Pattern.DOTALL));
+    private static final CP SHOW_VERSION = new CP("show version\n", Pattern.compile(".*GNU.*", Pattern.DOTALL));
```
jerboaa, line 275: Shall we use `gdb` over `GNU`? We expect a `gdb` version after all.

jerboaa (approved):

testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
zakkak: It should be reasonably safe to use `gdb`, the reason I chose GNU is because GDB is licensed under GPL and thus it's required to show a notice about the license in interactive mode:

> If the program does terminal interaction, make it output a short notice like this when it starts in an interactive mode:

Checking https://www.gnu.org/licenses/gpl-3.0.html though I see that it's not mandatory to mention the license itself so `GNU` might not be the best choice after all... `NO WARRANTY` on the other hand... :)

In any case, let's go with gdb and hope that maintainers keep it in the version string in all distributions.

jerboaa (approved):

PR #162, opened 2023-06-05
Fix pattern for debug symbols smoke test when using JDK 21

zakkak:
Fixes issue seen in https://github.com/graalvm/mandrel/actions/runs/5172441559/jobs/9317239703#step:11:18618

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
@@ -145,7 +145,7 @@ public enum GDBSession {
                                         ".*File debug_symbols_smoke/ClassA.java:.*" +
                                                 "java.lang.String \\*debug_symbols_smoke.ClassA::toString\\(\\).*" +
                                                 "File debug_symbols_smoke/Main.java:.*" +
-                                                "void debug_symbols_smoke.Main\\$\\$Lambda\\$.*::accept\\(java.lang.Object\\*\\).*" +
+                                                "void debug_symbols_smoke.Main\\$\\$Lambda.*::accept\\(java.lang.Object\\*\\).*" +
                                                 "void debug_symbols_smoke.Main::lambda\\$thisIsTheEnd\\$0\\(java.io.ByteArrayOutputStream\\*, debug_symbols_smoke.ClassA\\*\\).*" +
                                                 "void debug_symbols_smoke.Main::main\\(java.lang.String\\[\\]\\*\\).*" +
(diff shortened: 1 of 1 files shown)


testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
```diff
@@ -134,6 +134,52 @@ public CP[] get(boolean inContainer) {
...
+                        SHOW_VERSION,
+                        new CP("run < ./test_data_small.txt\n",
+                                Pattern.compile(".*fdc7c50f390c145bc58a0bedbe5e6d2e35177ac73d12e2b23df149ce496a5572.*exited normally.*",
+                                        Pattern.DOTALL)),
+                        new CP("info functions .*smoke.*\n",
+                                Pattern.compile(
+                                        ".*File debug_symbols_smoke/ClassA.java:.*" +
+                                                "java.lang.String \\*debug_symbols_smoke.ClassA::toString\\(\\).*" +
+                                                "File debug_symbols_smoke/Main.java:.*" +
+                                                "void debug_symbols_smoke.Main\\$\\$Lambda/0x.*::accept\\(java.lang.Object\\*\\).*" +
```
zakkak, line 148: This is the line that got changed. Lambdas are named differently in JDK 21.

testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
```diff
@@ -134,6 +134,52 @@ public CP[] get(boolean inContainer) {
                                 Pattern.compile(".*ClassA\\(int myNumber, String myString\\).*",
                                         Pattern.DOTALL)),
                 };
+            } else if (UsedVersion.jdkFeature(inContainer) == 21) {
```
jerboaa, line 137: `>= 21`?

jerboaa (commented):
Could we consider a more lose pattern that matches more of those `accept` methods?

testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
zakkak: You mean in order to match both JDK < 21 and JDK >= 21 lambda names. Sure it would.

zakkak:
> Could we consider a more lose pattern that matches more of those `accept` methods?

Yes, that would probably be a better strategy

testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
jerboaa: Yes, that's the idea.

jerboaa (approved):
LGTM.

Issue #160, opened 2023-06-01
DebugSymbolsTest.debugSymbolsQuarkus fails with latest graal/mandrel main


jerboaa:
Fails locally with:

```
java.net.ConnectException: Connection refused
	at java.base/sun.nio.ch.Net.pollConnect(Native Method)
	at java.base/sun.nio.ch.Net.pollConnectNow(Net.java:672)
	at java.base/sun.nio.ch.NioSocketImpl.timedFinishConnect(NioSocketImpl.java:542)
	at java.base/sun.nio.ch.NioSocketImpl.connect(NioSocketImpl.java:597)
	at java.base/java.net.Socket.connect(Socket.java:633)
	at java.base/sun.net.NetworkClient.doConnect(NetworkClient.java:178)
	at java.base/sun.net.www.http.HttpClient.openServer(HttpClient.java:531)
	at java.base/sun.net.www.http.HttpClient.openServer(HttpClient.java:636)
	at java.base/sun.net.www.http.HttpClient.<init>(HttpClient.java:279)
	at java.base/sun.net.www.http.HttpClient.New(HttpClient.java:384)
	at java.base/sun.net.www.http.HttpClient.New(HttpClient.java:406)
	at java.base/sun.net.www.protocol.http.HttpURLConnection.getNewHttpClient(HttpURLConnection.java:1309)
	at java.base/sun.net.www.protocol.http.HttpURLConnection.plainConnect0(HttpURLConnection.java:1242)
	at java.base/sun.net.www.protocol.http.HttpURLConnection.plainConnect(HttpURLConnection.java:1128)
	at java.base/sun.net.www.protocol.http.HttpURLConnection.connect(HttpURLConnection.java:1057)
	at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1665)
	at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1589)
	at org.graalvm.tests.integration.utils.WebpageTester.getUrlContents(WebpageTester.java:99)
	at org.graalvm.tests.integration.DebugSymbolsTest.lambda$carryOutGDBSession$3(DebugSymbolsTest.java:395)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[...]

PR #163, opened 2023-06-06
Increase timeout for DebugSymbolsTest#debugSymbolsQuarkus

zakkak:
Increases timeout for setting breakpoint in `DebugSymbolsTest#debugSymbolsQuarkus`.

Avoids failures due to https://github.com/Karm/mandrel-integration-tests/issues/160

The timeout should be revisited once we conclude on whether such a delay on loading the symbols is acceptable or not.

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
@@ -237,7 +237,10 @@ public enum GDBSession {
                     SHOW_VERSION,
                     new CP("b ConfigTestController.java:33\n",
                             Pattern.compile(".*Breakpoint 1 at .*: file com/example/quarkus/config/ConfigTestController.java, line 33.*",
-                                    Pattern.DOTALL)),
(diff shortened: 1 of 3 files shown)


testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
```diff
@@ -237,7 +237,8 @@ public CP[] get(boolean inContainer) {
                     SHOW_VERSION,
                     new CP("b ConfigTestController.java:33\n",
                             Pattern.compile(".*Breakpoint 1 at .*: file com/example/quarkus/config/ConfigTestController.java, line 33.*",
-                                    Pattern.DOTALL)),
+                                    Pattern.DOTALL),
+                            5 * 60), //FIXME: The timeout should be revisited once we conclude on whether such a delay on loading the symbols is acceptable or not.
```
jerboaa, line 241: Please mention #160 in this comment.

jerboaa (approved):
OK. We need to reduce test noise as much as we can.

jerboaa (approved):

Issue #173, opened 2023-07-26
gdb `show version` times out with JDK 21+32 and latest `graal/master`


zakkak:
CI runs fail due to `show version` command timing out (see https://github.com/Karm/mandrel-integration-tests/issues/168#issuecomment-1651134484)

Augmenting mandrel-integration-tests with https://github.com/Karm/mandrel-integration-tests/pull/172, I see that `show version` takes ~15s:

```
show version


Command took 15001 ms
```

What's really interesting is that the next command which is setting a breakpoint and should be the one taking ~15s only takes 1s

```
b ConfigTestController.java:33


Command took 1000 ms

(gdb) Breakpoint 1 at 0x40e806: file com/example/quarkus/config/ConfigTestController.java, line 33.
```

zakkak:
OK, so the 15s are actually spent in the first command `gdb ./target/quarkus-runner` but since the buffer matches the expected pattern before the completion of the command `waitForBufferToMatch` returns faster and then it appears like `show version` is taking these 15s.

PR #174, opened 2023-07-26
Improve handling of gdb commands

zakkak:
Instead of waiting till the gdb output matches the expected command or timeouts, wait for the command to complete (or timeout), i.e. for the prompt to appear. This way if a command prints the expected output but doesn't complete, it's execution time is not being counted by the next `waitForBufferToMatch` invocation, additionally if the command completes but doesn't return the expected result we don't have to wait for the timeout. Furthermore, we can enhance the logging to see if the command timed out or completed but without matching the expected output.

To ensure the prompt appears consistently across gdb versions after every command we use the GDB/MI mode, i.e. the "--interpreter=mi" option.

Closes https://github.com/Karm/mandrel-integration-tests/issues/173

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
@@ -135,7 +135,7 @@ public class DebugSymbolsTest {
             // We should somehow capture this semantically in an Enum or something. This is fragile...
             builderRoutine(app.buildAndRunCmds.cmds.length - 2, app, report, cn, mn, appDir, processLog, null, switches);
 
-            final ProcessBuilder processBuilder = new ProcessBuilder(getRunCommand("gdb", "./target/debug-symbols-smoke"));
+            final ProcessBuilder processBuilder = new ProcessBuilder(getRunCommand("gdb", "--interpreter=mi", "./target/debug-symbols-smoke"));
             final Map<String, String> envA = processBuilder.environment();
             envA.put("PATH", System.getenv("PATH"));
             processBuilder.directory(appDir)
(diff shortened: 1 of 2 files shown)


zakkak:
As a side effect, GDB/MI also gives more informative output in case of failures, e.g. in https://github.com/Karm/mandrel-integration-tests/actions/runs/5669589091/job/15362638665?pr=174#step:5:1268

it shows:

```
Command 'bt' did not match the expected pattern '.*at org/acme/vertx/Fruit.java:48.*'.
Output was:
"bt
""Selected thread is running.
"^error,msg="Selected thread is running."(gdb)  ==> expected: <true> but was: <false>
```

Which is an indication that the breakpoint was not hit maybe because of the race condition described in https://github.com/Karm/mandrel-integration-tests/blob/5eace8d7031a241d3fb9f4a2b64cb083a8556198/testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java#L405

(I am rerunning the failed job to see if the issue is persistent)

Karm:
My only concern is that it breaks on older gdb, but lets move forward and merge and we can attend to that later as I don't have any apparent failure at hand now.

zakkak:
> My only concern is that it breaks on older gdb, but lets move forward and merge and we can attend to that later as I don't have any apparent failure at hand now.

There is the option to make it more backwards friendly by passing `--interpreter=mi2`:

```
mi

    The newest GDB/MI interface (currently mi3). Used primarily by programs wishing to use GDB as a backend for a debugger GUI or an IDE. For more information, see [The GDB/MI Interface](https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI.html#GDB_002fMI).
mi3

    The GDB/MI interface introduced in GDB 9.1.
mi2

    The GDB/MI interface introduced in GDB 6.0.
mi1

    The GDB/MI interface introduced in GDB 5.1.
```

If you see any failure let me know on which gdb version it was.

PR #176, opened 2023-08-08
Smart getProperty(... was too smart with container runtime property

Karm:
Regression caused by fed26dd, happens for some *podman* settings on command line, e.g. 
```
mvn clean verify -Ptestsuite-builder-image -Dquarkus.version=3.2.3.Final -Dquarkus.native.container-runtime=podman
```

must work too. Not just QUARKUS_NATIVE_CONTAINER_RUNTIME env prop.


jerboaa (approved):
LGTM, but this transformation of alternative names takes a mental toll :-/ I've once run into this weirdness myself.

PR #171, opened 2023-07-24
Update Quarkus and Mandrel versions in CI workflows

zakkak:

PR #177, opened 2023-08-09
aarch64 VMs: Slower containers, GDB timeouts, Postgress 10 not for v8 aarch64

Karm:
e.g. 
```
FAIL_ON_PERF_REGRESSION=false mvn clean verify -Ptestsuite-builder-image -Dtest=DebugSymbolsTest#debugSymbolsQuarkusContainer -Dquarkus.native.builder-image=quay.io/quarkus/ubi-quarkus-mandrel-builder-image:jdk-17 -Dquarkus.version=3.2.3.Final
```

Also a follow up to #175


testsuite/src/it/java/org/graalvm/tests/integration/utils/GDBSession.java
```diff
@@ -233,12 +235,13 @@ public CP[] get(boolean inContainer) {
     DEBUG_QUARKUS_FULL_MICROPROFILE {
         @Override
         public CP[] get(boolean inContainer) {
+            // The huge timeout is needed because it takes a very long time to set a breakpoint, even after: https://github.com/graalvm/mandrel/pull/545
+            final int increasedTimeout = (UsedVersion.getVersion(inContainer).compareTo(Version.create(23, 0, 0)) >= 0) ? 60 : DEFAULT_TIMEOUT_SECONDS;
```
jerboaa, line 239: Could you please do me a favour and measure the time it takes for the breakpoint to hit with something like this:

```
gdb -ex 'set breakpoint pending on' -ex 'set debuginfod enabled off' -ex "set \$starttime=$(date +%s.%N)" -ex 'b ConfigTestController.java:33' -ex 'shell echo set \$endtime=$(date +%s.%N) > ~/gdbtmp.txt' -ex 'source ~/gdbtmp.txt' -ex 'print $endtime-$starttime'  quarkus-runner
```

I'd like to see which numbers we get on various systems.

testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
```diff
@@ -410,7 +411,10 @@ public static void carryOutGDBSession(StringBuffer stringBuffer, GDBSession gdbS
                                 Thread.sleep(100);
                             } while (gotoException != null && tryCount > 0);
                             if (gotoException != null) {
-                                fail("Unexpected GOTO failure: ", gotoException);
+                                errorQueue.add("Unexpected GOTO URL " + cp.c.split("URL ")[1] + " failure: " + gotoException.fillInStackTrace().toString());
+                            }
+                            if (MAX_GOTO_TRIES - tryCount < MAX_GOTO_TRIES - 2 && Commands.FAIL_ON_PERF_REGRESSION) {
```
jerboaa, line 416: Why this magic `- 2`? How about simplifying this condition to `tryCount > 2` and replace `2` with a constant?

jerboaa (commented):
A few questions :)

testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
Karm: Makes sense, Will amend. 

I am not sure what is reasonable here, hence magic numbers experimentation.
What would you suggest? 
The breakpoint should be set already at that time (slow). So why hitting it is so much slower now than before? With 100ms we are already at 100x :-D of what it would take to serve the page without debug session going on.

My point is not to nitpick on milliseconds. I am genuinely worried that if a giant timeout of 200m is not enough and we need 500ms+ here for such a test app, it actually might become unusable for a production, big app where attaching to a live system and clicking through app would cause browser to time out before hitting the breakpoint....

Karm:
@jerboaa 

1) gdb numbers across our various tests systems:
WIP

2) Amending the timeouts:
The latest commit takes the GOTO URL for hitting web app url during gdb session for a different approach using timeout, not trials. I also removed the FAIL_ON_PERF_REGRESSION check.  I.e. less magic I hope.

Karm:
The #171 has an incorrect exclusion matrix, e.g. 

```
Error: [ERROR] 	[error]: Build step io.quarkus.deployment.pkg.steps.NativeImageBuildStep#build threw an exception: java.lang.IllegalStateException: Out of date version of GraalVM detected: native-image 17.0.8 2023-07-18. Quarkus currently supports 22.3.0. Please upgrade GraalVM to this version.
```
trying to run Q 2.16 with M 23 on Windows.

jerboaa:
> trying to run Q 2.16 with M 23 on Windows.

OK. Let's fix this first and then we have a better idea about the state of this PR. Sound good?

testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
jerboaa: Not sure if debugging and attaching to production workloads is the primary goal of the debug info work. IMO it's more for expert developers trying to fix a bug in native code (outside production).

Karm:
Hmm, looks like we are also hit by something quite unrelated here;
```
 Error:    RuntimesSmokeTest.quarkusEncodingIssues:212->testRuntime:166 build-and-run.log log should not contain error or warning lines that are not whitelisted. See /home/runner/work/mandrel-integration-tests/mandrel-integration-tests/ts/testsuite/target/archived-logs/org.graalvm.tests.integration.RuntimesSmokeTest/quarkusEncodingIssues/build-and-run.log and check these offending lines: 
Warning:  Checksum validation failed, expected 2811ba27a71a8bda0602161ffe2f6e1429da8068 but is 36257165a0945753efb3f9d473d86c6f4c6c6f6e from google-maven-central for https://maven-central.storage-download.googleapis.com/maven2/org/jboss/arquillian/arquillian-bom/1.7.0.Final/arquillian-bom-1.7.0.Final.pom
Warning:  Could not validate integrity of download from https://maven-central.storage-download.googleapis.com/maven2/org/jboss/arquillian/arquillian-bom/1.7.0.Final/arquillian-bom-1.7.0.Final.pom
```

I'll bump Q version...

Karm:
This:

```
    strategy:
      fail-fast: false
      matrix:
        os: [ ubuntu-20.04 ]
        include:
          - quarkus-version: '2.13.8.Final'
            mandrel-builder-image: '22.3-java17'
          - quarkus-version: '2.16.8.Final'
            mandrel-builder-image: '22.3-java17'
          - quarkus-version: '3.2.2.Final'
            mandrel-builder-image: 'jdk-17'
```

Makes only the last line run.  I will revert to the previous older matrix format.

Karm:
Fighting GHA aside, I can confirm that the 23.0 passes GDB tests also on slow aarch64 VMs now with this branch.

.github/workflows/builder_image_tests.yml
```diff
@@ -25,12 +26,16 @@ jobs:
       matrix:
         os: [ ubuntu-20.04 ]
         include:
+          - quarkus-version: '2.7.7.Final'
+            mandrel-builder-image: '21.3-java17'
```
jerboaa, line 30: Those two lines can probably go. It's EOL.

jerboaa (approved):
Seems OK. Consider removing 21.3 + 2.7 Q combinations.

.github/workflows/builder_image_tests.yml
Karm: as long as you can pick it here... https://code.quarkus.redhat.com/  (top dropdown)

jerboaa:
Some CI failures don't seem to cope well with extra output. Example from [here](https://github.com/Karm/mandrel-integration-tests/actions/runs/5833583288/job/15821335331?pr=177#step:8:2556):

```
There were errors in the GDB session. Note that commands in the session might depend on each other. Errors: 
Command 'b ConfigTestController.java:33' did not match the expected pattern '.*Breakpoint 1 at .*: file com/example/quarkus/config/ConfigTestController.java, line 33.*'.
Output was:
"b ConfigTestController.java:33
", 
Command 'run&' did not match the expected pattern '.*Installed features:.*'.
Output was:
Breakpoint 1 at 0x40e766: file com/example/quarkus/config/ConfigTestController.java, line 33.
breakpoint-created,bkpt={number="1",type="breakpoint",disp="keep",enabled="y",addr="0x000000000040e766",func="com.example.quarkus.config.ConfigTestController::getLookupConfigValue()",file="com/example/quarkus/config/ConfigTestController.java",fullname="/home/runner/work/mandrel-integration-tests/mandrel-integration-tests/ts/apps/quarkus-full-microprofile/target/quarkus-native-image-source-jar/sources/com/example/quarkus/config/ConfigTestController.java",line="33",thread-groups=["i1"],times="0",original-location="ConfigTestController.java:33"}^done(gdb) "run&
"Starting program: /home/runner/work/mandrel-integration-tests/mandrel-integration-tests/ts/apps/quarkus-full-microprofile/target/quarkus-runner 
thread-group-started,id="i1",pid="16284"thread-created,id="1",group-id="i1", 
Unexpected GOTO URL http://localhost:8080/data/config/lookup failure in 60034ms, see: java.net.ConnectException: Connection refused ==> expected: <true> but was: <false>
	at org.graalvm.tests.integration.DebugSymbolsTest.carryOutGDBSession(DebugSymbolsTest.java:452)
	at org.graalvm.tests.integration.DebugSymbolsTest.debugSymbolsQuarkus(DebugSymbolsTest.java:263)
```

Line is there:

```
Breakpoint 1 at 0x40e766: file com/example/quarkus/config/ConfigTestController.java, line 33.
```

but has extra output...

Karm:
@jerboaa It is timing on GHA.

```
$ export JAVA_HOME=/home/karm/workspaceRH/mandrel-release/mandrel-java20-23.0.1.2-Final;export GRAALVM_HOME=${JAVA_HOME};export PATH=${JAVA_HOME}/bin:${PATH}
$ mvn clean verify -Ptestsuite -DincludeTags=reproducers,perfcheck,runtimes -Dquarkus.version=3.2.3.Final
<SNIP>
[INFO] Native image integration TS ........................ SUCCESS [  0.105 s]
[INFO] testsuite .......................................... SUCCESS [27:34 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  27:35 min
[INFO] Finished at: 2023-08-11T22:07:56+02:00
[INFO] ------------------------------------------------------------------------
```

The question is how to make it more resilient to that. Apparently the crude timeout per gdb command ain't enough. The text appears in the StringBuffer but is never checked again...?

Karm:
Re-running those two failed...

Karm:
I am not giving up : ) 

When I use my local Ubuntu VM and set it to 2 vCPU and 8G RAM, I can reproduce the issue:

```
INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   DebugSymbolsTest.debugSymbolsQuarkus:263->carryOutGDBSession:452 There were errors in the GDB session. Note that commands in the session might depend on each other. Errors: 
Command 'run&' did not match the expected pattern '.*Installed features:.*'.
Output was:
"run&
"Starting program: /home/karm/mandrel-integration-tests/apps/quarkus-full-microprofile/target/quarkus-runner 
thread-group-started,id="i1",pid="6539"thread-created,id="1",group-id="i1"library-loaded,id="/lib64/ld-linux-x86-64.so.2",target-name="/lib64/ld-linux-x86-64.so.2",host-name="/lib64/ld-linux-x86-64.so.2",symbols-loaded="0",thread-group="i1",ranges=[{from="0x00007ffff7fc5090",to="0x00007ffff7fee335"}]^running*running,thread-id="all"(gdb) library-loaded,id="/lib/x86_64-linux-gnu/libz.so.1",target-name="/lib/x86_64-linux-gnu/libz.so.1",host-name="/lib/x86_64-linux-gnu/libz.so.1",symbols-loaded="0",thread-group="i1",ranges=[{from="0x00007ffff7f92280",to="0x00007ffff7fa2c14"}]library-loaded,id="/lib/x86_64-linux-gnu/libc.so.6",target-name="/lib/x86_64-linux-gnu/libc.so.6",host-name="/lib/x86_64-linux-gnu/libc.so.6",symbols-loaded="0",thread-group="i1",ranges=[{from="0x00007ffff7c28700",to="0x00007ffff7dbaabd"}][Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
 ==> expected: <true> but was: <false>
[INFO] 
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
```

When I run it locally on my CentOS 8 Stream workstation (12 cores, 32G RAM), I cannot reproduce it any more.

So I will keep poking at it until it's reliable.

Karm:
I think the problem is that `run&` command immediately returns and that different gdb versions decide differently when to spit output, see the key difference beginning line 84:

https://editor.mergely.com/u0uC749D

On my fast CentOS system on the right, the output fed immediately after the `run&` command contains the expected string about Installed features, marking Quarkus is started.
On the left panel, the somewhat feeble 2 vCPU Ubuntu VM, with a much different gdb version, does not spit any such output until a next command, `bt` is called. Hence the parsing fails.

I will try to jam some no-op command in the session to see if we could make it spit all the output sooner.

Karm:
As I play with it, I'd put out there an idea that intermittently, clearly due to some race condition, debugged Quarkus app does not get into a correct state where it would be able to serve webpages at all, i.e. it keeps turning Connection Refused even minutes after 'run&' command executed.  Sometimes.

testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
```diff
@@ -161,12 +167,14 @@ public void debugSymbolsSmokeGDB(TestInfo testInfo) throws IOException, Interrup
             try (BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(process.getOutputStream()))) {
                 Logs.appendlnSection(report, appDir.getAbsolutePath());
                 Logs.appendln(report, String.join(" ", processBuilder.command()));
+                final long increasedTimeoutMs = (UsedVersion.getVersion(false)
+                        .compareTo(Version.create(23, 0, 0)) >= 0) ? CMD_LONG_TIMEOUT_MS : CMD_DEFAULT_TIMEOUT_MS;
                 boolean result = waitForBufferToMatch(report, stringBuffer,
                         Pattern.compile(".*Reading symbols from.*", Pattern.DOTALL),
-                        3000, 500, TimeUnit.MILLISECONDS);
+                        increasedTimeoutMs, 500, TimeUnit.MILLISECONDS); // Time unit s the same for timeout and sleep.
```
zakkak, line 174:
```suggestion
                        increasedTimeoutMs, 500, TimeUnit.MILLISECONDS); // Time unit is the same for timeout and sleep.
```

testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
```diff
@@ -231,9 +239,11 @@ public void debugSymbolsQuarkus(TestInfo testInfo) throws IOException, Interrupt
 
             Logs.appendlnSection(report, appDir.getAbsolutePath());
             Logs.appendln(report, String.join(" ", processBuilder.command()));
+            final long increasedTimeoutMs = (UsedVersion.getVersion(false)
+                    .compareTo(Version.create(23, 0, 0)) >= 0) ? CMD_LONG_TIMEOUT_MS : CMD_DEFAULT_TIMEOUT_MS;
             boolean result = waitForBufferToMatch(report, stringBuffer,
                     Pattern.compile(".*Reading symbols from.*", Pattern.DOTALL),
-                    60000, 500, TimeUnit.MILLISECONDS);
+                    increasedTimeoutMs, 500, TimeUnit.MILLISECONDS); // Time unit s the same for timeout and sleep.
```
zakkak, line 246:
```suggestion
                    increasedTimeoutMs, 500, TimeUnit.MILLISECONDS); // Time unit is the same for timeout and sleep.
```

testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
```diff
@@ -341,13 +351,14 @@ public void debugSymbolsQuarkusContainer(TestInfo testInfo) throws IOException,
             try (BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(gdbProcess.getOutputStream()))) {
                 Logs.appendlnSection(report, appDir.getAbsolutePath());
                 Logs.appendln(report, String.join(" ", processBuilder.command()));
+                final long increasedTimeoutMs = (UsedVersion.getVersion(true)
+                        .compareTo(Version.create(23, 0, 0)) >= 0) ? CMD_LONG_TIMEOUT_MS : CMD_DEFAULT_TIMEOUT_MS;
                 boolean result = waitForBufferToMatch(report, stringBuffer,
                         Pattern.compile(".*Reading symbols from.*", Pattern.DOTALL),
-                        3000, 500, TimeUnit.MILLISECONDS);
+                        increasedTimeoutMs, 500, TimeUnit.MILLISECONDS); // Time unit s the same for timeout and sleep.
```
zakkak, line 358:
```suggestion
                        increasedTimeoutMs, 500, TimeUnit.MILLISECONDS); // Time unit is the same for timeout and sleep.
```

testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
zakkak, line 174: Once concern I have with having a global timeout is that we might miss slow downs that affect different tests. E.g., at the moment the only affected test (that requires much longer time to start) is the quarkus-microprofile, by increasing the timeout for all tests we might miss the opportunity to notice a slow down in other tests.

Karm: That is false, it's all Quarkus tests, i.e. both Microprofile and Vert.x one.
+1 for making it more fine grained when we get it working at least with what we have.

Karm:
@zakkak  I don't know... 

```
Unexpected GOTO URL http://localhost:8080/data/config/lookup failure in 300014ms, see: java.net.SocketTimeoutException: Connect timed out
```

zakkak:
Looking at the logs I see that `gdb` starts fine in 18505ms. Then spents 500ms to show the version, Then 54012ms to set the breakpoint and another 500ms to `run` the application and interestingly it stops at the breakpoint, which is an indication that the application and gdb behave as expected !?

testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
```diff
@@ -397,26 +414,33 @@ public static void carryOutGDBSession(StringBuffer stringBuffer, GDBSession gdbS
                                     final String url = cp.c.split("URL ")[1];
                                     final String content = WebpageTester.getUrlContents(url);
                                     if (!cp.p.matcher(content).matches()) {
-                                        errorQueue.add("Content of URL " + url + " should have matched regexp " + cp.p.pattern() + " but it was this: " + content);
+                                        errorQueue.add("Content of URL " + url + " should have matched regexp " + cp.p.pattern() +
+                                                " but it was this: " + content);
                                     }
+                                    gotoException = null;
```
zakkak, line 420:
```suggestion
```

testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
```diff
@@ -397,26 +414,33 @@ public static void carryOutGDBSession(StringBuffer stringBuffer, GDBSession gdbS
...
                                 }
                             };
-                            int tryCount = MAX_GOTO_TRIES;
+                            final long gotoURLStart = System.currentTimeMillis();
+                            final long timeoutMs = (UsedVersion.getVersion(inContainer)
+                                    .compareTo(Version.create(23, 0, 0)) >= 0) ? LONG_GOTO_URL_TIMEOUT_MS : GOTO_URL_TIMEOUT_MS;
+                            long duration = 0;
                             do {
-                                tryCount--;
                                 esvc.submit(webRequest);
```
zakkak, line 430:
```suggestion
                                gotoException = null;
                                esvc.submit(webRequest);
```

Reset `gotoException` from the `main` thread and only allow it to be `set` by the request thread. This way we can't have multiple threads running at the same time and altering the variable.

Right now we can have main spawn a request `T1` which will fail, then `main` will spawn a new request `T2` while keeping the `gotoException` set. Assuming `T2` succeeds, it will hit the breakpoint which takes more than `GOTO_URL_SLEEP_MS`. As a result `T2` is still active but hasn't reset `gotoException`, `main` doesn't wait for it and keeps spawning `T3`, `T4`, etc. which all fail and constantly set `gotoException` possibly overriding `T2`'s reset when it completes.

zakkak (changes requested):

Karm:
@zakkak Is it a Firefox glitch or are all check finally green? :-) 

Let me know if you are O.K. with the solution. I have a strong feeling it could be drastically simplified.  On the other hand, I am happy I finally know what's going on...

testsuite/src/it/java/org/graalvm/tests/integration/DebugSymbolsTest.java
Karm: You are correct. It was exactly this problem.

testsuite/src/it/java/org/graalvm/tests/integration/utils/versions/UsedVersion.java
```diff
@@ -149,14 +149,16 @@ static MVersion parse(List<String> lines) {
...
                 VersionBuilder builder = new VersionBuilder();
-                return builder.jdkUsesSysLibs(false /* not implemented */)
-                       .beta("" /* not implemented */)
-                       .jdkFeature(v.feature())
-                       .jdkInterim(v.interim())
-                       .jdkUpdate(v.update())
-                       .version(vers)
-                       .build();
+                return builder
+                        .jdkUsesSysLibs(lastLine.contains("-LTS"))
```
jerboaa, line 155: I don't think this will be necessarily true. `LTS` in the version output `!=` whether or not system libs are being used. It seems safe to assume that lastest releases no longer use system libs. Not sure if this condition still has any meaningful value.

jerboaa (approved):
Seems OK to me and better to what we have live currently.

zakkak (approved):
LGTM, thanks @Karm

testsuite/src/it/java/org/graalvm/tests/integration/utils/versions/UsedVersion.java
Karm: It is used to distinguish Mandrel based on Red Hat build of OpenJDK that has this different sys libs behavior from other distros.

Karm: @jerboaa i.e. with "LTS" in there, we skip doing hashsum checks due to slightly different anti aliasing on fonts...
https://github.com/Karm/mandrel-integration-tests/blob/master/testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java#L583

While I am writing this, I wonder if it's something to be revisited again...  (not in this PR though I guess)

PR #191, opened 2023-09-07
Improves aarch64 compatibility (hyperfoil), plus small fixes

Karm:
Check if executable exists first

amd64/arm64 Hyperfoil, thresholds props

Fiddle with CPU boost only where supported

Cleanup imageio cache font files, Whitelist podman

Relaxes <= 0 to < 0 for measurements as diff of size 0 is valid

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
@@ -228,13 +228,13 @@ public enum BuildAndRunCmds {
                     "-XX:+FlightRecorder",
                     "-XX:StartFlightRecording=settings=" + BASE_DIR + File.separator + "apps" + File.separator + "jfr-native-image-performance/jfr-perf.jfc,filename=logs/flight-native.jfr",
                     "-XX:FlightRecorderLogging=jfr"},
-            new String[]{CONTAINER_RUNTIME, "run", "--name", ContainerNames.HYPERFOIL.name, "--rm", "--network=host", "quay.io/hyperfoil/hyperfoil:0.25.2", "standalone"}
+            new String[]{CONTAINER_RUNTIME, "run", "--name", ContainerNames.HYPERFOIL.name, "--rm", "--network=host", "quay.io/karmkarm/hyperfoil:0.25.2", "standalone"}
     }),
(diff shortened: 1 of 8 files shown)


jerboaa (approved):
Seems fine.

testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
```diff
@@ -250,7 +250,7 @@ public enum BuildAndRunCmds {
                     "--name", ContainerNames.JFR_PERFORMANCE_BUILDER_IMAGE.name, "jfr-performance-app", "-XX:+FlightRecorder",
                     "-XX:StartFlightRecording=settings=/work/jfr-perf.jfc,filename=/tmp/flight-native.jfr",
                     "-XX:FlightRecorderLogging=jfr"},
-            new String[]{CONTAINER_RUNTIME, "run", "--name", ContainerNames.HYPERFOIL.name, "--rm", "--network=host", "quay.io/hyperfoil/hyperfoil:0.25.2", "standalone"}
+            new String[]{CONTAINER_RUNTIME, "run", "--name", ContainerNames.HYPERFOIL.name, "--rm", "--network=host", "quay.io/karmkarm/hyperfoil:0.25.2", "standalone"}
```
zakkak, line 253: I guess this is needed because the original hyperfoil image is not aarch64 compatible right? I wonder if it would be better to [host these images in this github repo](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) instead of quay.io to have all things in a single place.

zakkak (approved):
LGTM

testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java
Karm: @zakkak The official quay.io/hyperfoil/hyperfoil is amd64 only. They build the image with Maven as a part of the standard build process, followed by a docker push.

I built my own amd64 and aarch64 images and created a multiarch image, pushed to my account.

> better to [host these images in this github repo](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) instead of quay.io to have all things in a single place.

I will take a look. It's good that you would have push rights too.

Issue #186, opened 2023-09-07
[23.1] imageioAWTTest integration test fails with experimental warning


jerboaa:
It fails with:

```
imageioAWTTest{TestInfo}  Time elapsed: 55.132 s  <<< FAILURE!
org.opentest4j.AssertionFailedError:
build-and-run.log log should not contain error or warning lines that are not whitelisted. See /disk/graal/upstream-sources/mandrel-integration-tests/testsuite/target/archived-logs/org.graalvm.tests.integration.AppReproducersTest/imageioAWTTest/build-and-run.log and check these offending lines:
Warning: The option '-H:IncludeResources=Grace_M._Hopper.jp2,MyFreeMono.ttf,MyFreeSerif.ttf' is experimental and must be enabled via '-H:+UnlockExperimentalVMOptions' in the future.
Warning: Please re-evaluate whether any experimental option is required, and either remove or unlock it. The build output lists all active experimental options, including where they come from and possible alternatives. If you think an experimental option should be considered as stable, please file an issue. ==> expected: <true> but was: <false>
        at org.graalvm.tests.integration.AppReproducersTest.imageioAWT(AppReproducersTest.java:643)
        at org.graalvm.tests.integration.AppReproducersTest.imageioAWTTest(AppReproducersTest.java:498)
```

jerboaa:
Related #184

jerboaa:
As per https://github.com/oracle/graal/issues/7354 `-H:IncludeResources` shouldn't be used. See https://github.com/oracle/graal/issues/7354#issuecomment-1707965060 for a replacement.

jerboaa:
Reproducer:

```
export JAVA_HOME=/path/to/mandrel
export GRAALVM_HOME=$JAVA_HOME
export PATH=$JAVA_HOME/bin:$PATH
mvn clean verify -Dquarkus.version=999-SNAPSHOT -Ptestsuite -Dtest=AppReproducersTest#imageioAWTTest
```

PR #194, opened 2023-09-07
Fix imageio test for 23.1+

jerboaa:
Resources don't need to get explicitly added with the CLI option, since the java run with the agent, will generate appropriate resource-config.json files which are sufficient for the resources to get added.

By removing `-H:IncludeResources` we no longer get the experimental VM warnings on 23.1+ (#186).


Karm:
'''
Error:  Failures: 
Error:    DebugSymbolsTest.debugSymbolsQuarkus:222 Quarkus executable does not exist. Compilation failed. Check the logs. ==> expected: <true> but was: <false>
'''

Looks serious. I cannot pull the logs now, let's have a look at .log file.

Karm (approved):
Thx.

PR #200, opened 2023-09-11
Fix version test exclusion

jerboaa:
The earlier fix missed to exclude 23.1 as the condition was 23.1.0.0-dev<sha> <= 23.1 which is true (wrong). What we actually want is to run it up to and including to 23.0. Latest release is 23.0.1, so up to a max value of 23.0 doesn't work either, since that would amount to 23.0.1 <= 23.0.0 which is false. Use 23.0.99 as the comparison value instead.


jerboaa:
Fixes issues like these:
https://github.com/graalvm/mandrel/actions/runs/6144168401/job/16670211716#step:11:17929

zakkak (approved):

Issue #211, opened 2023-10-02
Possible unnecessary image builds in JFR perf test


roberttoyonaga:
Do we need to [rebuild the JFR perf test app for each test configuration](https://github.com/Karm/mandrel-integration-tests/blob/master/testsuite/src/it/java/org/graalvm/tests/integration/JFRTest.java#L322-L327)? Currently each app is being built twice (resulting in 4 native builds each time `jfrPerfTestRun` is called.) This results in a total of 8 image builds for all the JFR perf tests.
![image](https://github.com/Karm/mandrel-integration-tests/assets/35396843/e317032b-0813-4500-a909-dfaa93b62122)

Both test endpoints are included in each JFR app so there's no need to rebuild based on the endpoint we are hitting in the hyperfoil benchmark. I think the tree could look like this instead: 
![image](https://github.com/Karm/mandrel-integration-tests/assets/35396843/15d1d9c2-e795-493c-b806-6e8a3c09b066)

This should almost cut the test time in half because the build is probably the lengthiest part. What do you think? If you agree we can reduce the number of builds, I'm happy to submit a follow up PR.

jerboaa:
@Karm PTAL.

Karm:
Thx @roberttoyonaga  Lemme take a look. Given the way Mandrel is installed, I rarely run local & container at the same time, so it entirely escaped me that when you do it's rebuilt again.

Karm:
@roberttoyonaga  I can see the app being built twice for the same set of flags too:
```
$ grep -e "Command.* package " ./testsuite/target/archived-logs/org.graalvm.tests.integration.JFRTest/jfrPerfTest/build-and-run.log

Command: mvn clean package -Pnative -Dquarkus.version=3.2.5.Final -Dquarkus.native.monitoring=jfr \
-Dquarkus.native.additional-build-args=-H:+UnlockExperimentalVMOptions,-H:+SignalHandlerBasedExecutionSampler,-H:-UnlockExperimentalVMOptions \
-DfinalName=jfr-perf
Command: mvn clean package -Pnative -Dquarkus.version=3.2.5.Final -DfinalName=jfr-plaintext

Command: mvn clean package -Pnative -Dquarkus.version=3.2.5.Final -Dquarkus.native.monitoring=jfr \
[...]

PR #212, opened 2023-10-03
Reduce the number of builds during JFR perf test

roberttoyonaga:
### Summary
Related issue: https://github.com/Karm/mandrel-integration-tests/issues/211

This PR does a bit of rearranging to try to reduce duplicate JFR perf test app builds. I've just moved the build up a few levels so its done in `jfrPerfTestRun` and can be reused for multiple hyperfoil benchmarks. I also moved the cleanup code too.  And I did some minor renaming for consistency.


Karm:
Thank you @roberttoyonaga

PR #215, opened 2023-10-05
[CI] Update Quarkus versions

zakkak:


.github/workflows/local_tests.yml
```diff
@@ -37,33 +37,33 @@ jobs:
...
             mandrel-version: '23.0.1.2-Final'
             java-version: '20'
             os: 'windows-2022'
             timeout-minutes: 80
-          - quarkus-version: '3.2.3.Final'
+          - quarkus-version: '3.2.6.Final'
             mandrel-version: '23.0.1.2-Final'
             java-version: '20'
             os: 'ubuntu-20.04'
             timeout-minutes: 130
```
zakkak, line 60: Does it still make sense to keep JDK 20 tests around?

jerboaa: With JDK 21 released I'd say no.

jerboaa (approved):

.github/workflows/local_tests.yml
Karm: We don't have a JDK 21 release, so let's keep them there until we do :rofl:

zakkak:
> Re--running CI ad
> 
> ```
>   72 Terminating due to java.lang.OutOfMemoryError: Java heap space
>   73 The Native Image build process ran out of memory.
>   74 Please make sure your build system has more memory available.
> ```

Hmm, this is the second time we observe this... Are we setting `-J-Xmx5g`?

Karm:
@zakkak We are not IMHO.

zakkak:
> @zakkak We are not IMHO.

Might be worth trying it if issues persist.

Issue #220, opened 2023-11-13
RuntimesSmokeTest.quarkusFullMicroProfile/JFRTest.jfrPerfTest fail due to new Quarkus warnings with quarkus main (3.6)


jerboaa:
The `RuntimesSmokeTest.quarkusFullMicroProfile` test fails in CI with JDK 22 + GraalVM 24.0 with:

```
2023-11-13 02:28:31,843 WARN  [io.qua.run.con.DeprecatedRuntimePropertiesRecorder] (main) The 'quarkus.http.auth.form.redirect-after-login' config property is deprecated and should not be used anymore ==> expected: <true> but was: <false>
Error:    RuntimesSmokeTest.quarkusFullMicroProfile:192->testRuntime:157 build-and-run.log log should not contain error or warning lines that are not whitelisted. See /home/runner/work/mandrel/mandrel/mandrel-integration-tests/testsuite/target/archived-logs/org.graalvm.tests.integration.RuntimesSmokeTest/quarkusFullMicroProfile/build-and-run.log and check these offending lines: 
2023-11-13 03:03:04,462 WARN  [io.qua.run.con.DeprecatedRuntimePropertiesRecorder] (main) The 'quarkus.http.auth.form.redirect-after-login' config property is deprecated and should not be used anymore ==> expected: <true> but was: <false>
```

and

```
Error:  Failures: 
Error:    JFRTest.jfrPerfTest:176->jfrPerfTestRun:219 build-and-run.log log should not contain error or warning lines that are not whitelisted. See /home/runner/work/mandrel/mandrel/mandrel-integration-tests/testsuite/target/archived-logs/org.graalvm.tests.integration.JFRTest/jfrPerfTest/build-and-run.log and check these offending lines: 
2023-11-13 02:28:29,532 WARN  [io.qua.run.con.DeprecatedRuntimePropertiesRecorder] (main) The 'quarkus.http.auth.form.redirect-after-login' config property is deprecated and should not be used anymore
```

See: https://github.com/graalvm/mandrel/actions/runs/6844550940/job/18609261996#step:11:18840

jerboaa:
I have reproduced this locally with the quarkus microprofile app, but I have no idea where the property is coming from:

```
$ java -jar target/quarkus-app/quarkus-run.jar 
__  ____  __  _____   ___  __ ____  ______ 
 --/ __ \/ / / / _ | / _ \/ //_/ / / / __/ 
[...]

PR #222, opened 2023-11-15
Allow-list quarkus deprecation warning.

jerboaa:
Until https://github.com/quarkusio/quarkus/issues/37072 is fixed so as to reduce test noise.

Related: #220


jerboaa:
This also allow-lists the for the JFRPerf test since it re-uses the quarkus microprofile allow-listed lines in its setup.

michalvavrik:
FYI there is approved fix https://github.com/quarkusio/quarkus/pull/37130 (but not merged, so might be worth to wait a day to see how it goes)

jerboaa:
Even better, thanks!

jerboaa:
Fixed with https://github.com/quarkusio/quarkus/issues/37072 instead.

Issue #223, opened 2023-11-16
quarkus-full-microprofile fails on quarkus main


jerboaa:
```
Error:  Failed to execute goal on project quarkus-full-microprofile: Could not resolve dependencies for project org.graalvm.tests.integration:quarkus-full-microprofile:jar:1.0.0-SNAPSHOT: Could not find artifact io.quarkus:quarkus-smallrye-opentracing:jar:999-SNAPSHOT -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException
Error:  Tests run: 2, Failures: 0, Errors: 1, Skipped: 1, Time elapsed: 7.833 s <<< FAILURE! - in org.graalvm.tests.integration.RuntimesSmokeTest
Error:  quarkusFullMicroProfile{TestInfo}  Time elapsed: 7.833 s  <<< ERROR!
java.lang.IllegalArgumentException: Path to D:\a\mandrel\mandrel\apps\quarkus-full-microprofile\targetmust exist and regexp must nut be null.
	at org.graalvm.tests.integration.RuntimesSmokeTest.testRuntime(RuntimesSmokeTest.java:98)
	at org.graalvm.tests.integration.RuntimesSmokeTest.quarkusFullMicroProfile(RuntimesSmokeTest.java:192)
```

Looks related to a comment @zakkak made elsewhere:

`smallrye-opentracing` no longer being present https://github.com/quarkusio/quarkus/pull/36602 (native-tests.json needs adjustment, PR handling it https://github.com/quarkusio/quarkus/pull/37101)

Not sure how a fix would look like in integration tests. Added dependency for quarkus > someversion?

jerboaa:
See: https://github.com/graalvm/mandrel/actions/runs/6882513424/job/18721824498#step:10:7503

jerboaa:
@Karm Could you please look at this with some priority? This is generating a lot of noise.

zakkak:
Temporary work around to get the CI green again: https://github.com/Karm/mandrel-integration-tests/pull/224

zakkak:
> Hmm, how about we produce a profile that activates based on the quarkus version and adds the quarkiverse dependency for later versions?

[...]

PR #224, opened 2023-11-20
Temporary work around: Disable quarkus-full-microprofile dependent tests

zakkak:
Works around https://github.com/Karm/mandrel-integration-tests/issues/223 to get CI passing till it gets fixed.

One major drawback, other than the obvious one of skipping three good tests in Quarkus >= 3.6.0 is that we skip them for any SNAPSHOT build as well even for Quarkus < 3.6.0.


jerboaa (approved):
I don't like it, but it's better than having those noisy CI runs. Could you please file an issue which tracks fixing this properly?

PR #227, opened 2023-11-25
FullMicroprofile ported to new tracing API, MP6+ Q3.6+

### LINKED ISSUE #225: Adapt quarkus-full-microprofile based tests to work with Quarkus >= 3.6.0
Starting with Quarkus 3.6.0 `quarkus-smallrye-opentracing` is no longer a core extension. `quarkus-full-microprofile` depends on it and thus fails with Quarkus >= 3.6.0 (as observed in https://github.com/Karm/mandrel-integration-tests/issues/223)

https://github.com/Karm/mandrel-integration-tests/pull/224 disabled quarkus-full-microprofile based tests for Quarkus >= 3.6.0, which comes with the big drawback that we are no longer running 3 important tests.

@jerboaa suggested in https://github.com/Karm/mandrel-integration-tests/issues/223#issuecomment-1820603355:

> how about we produce a profile that activates based on the quarkus version and adds the quarkiverse dependency for later versions?

Karm:
Fixes #225 

```
$ mvn clean verify -Ptestsuite 
-Dtest=RuntimesSmokeTest#quarkusFullMicroProfile,DebugSymbolsTest#debugSymbolsQuarkus,PerfCheckTest#testQuarkusFullMicroProfile 
-Dquarkus.native.builder-image=quay.io/quarkus/ubi-quarkus-mandrel-builder-image:jdk-21 
-DincludeTags=runtimes,reproducers,perfcheck -Dquarkus.version=3.6.0 | tee log.log 

```

passes now.

### DIFF
--- a/apps/quarkus-full-microprofile/threshold.properties
+++ b/apps/quarkus-full-microprofile/threshold.properties
@@ -1,6 +1,6 @@
 linux.time.to.first.ok.request.threshold.ms=300
 linux.RSS.threshold.kB=120000
-linux.executable.size.threshold.kB=73500
+linux.executable.size.threshold.kB=79000
 windows.time.to.first.ok.request.threshold.ms=1537
 windows.RSS.threshold.kB=120000
-windows.executable.size.threshold.kB=75000
+windows.executable.size.threshold.kB=79000
--- a/apps/quarkus-full-microprofile/quarkus_3.6.x.patch
+++ b/apps/quarkus-full-microprofile/quarkus_3.6.x.patch
@@ -1,0 +1,247 @@
+diff --git a/apps/quarkus-full-microprofile/pom.xml b/apps/quarkus-full-microprofile/pom.xml
+index ce7df81..c047cd9 100644
+--- a/apps/quarkus-full-microprofile/pom.xml
++++ b/apps/quarkus-full-microprofile/pom.xml
+@@ -61,7 +61,7 @@
+         </dependency>
+         <dependency>
+             <groupId>io.quarkus</groupId>
+-            <artifactId>quarkus-smallrye-opentracing</artifactId>
++            <artifactId>quarkus-opentelemetry</artifactId>
+         </dependency>
+         <dependency>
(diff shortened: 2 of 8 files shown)


apps/quarkus-full-microprofile/threshold.properties
```diff
@@ -1,6 +1,6 @@
 linux.time.to.first.ok.request.threshold.ms=300
 linux.RSS.threshold.kB=120000
-linux.executable.size.threshold.kB=73500
+linux.executable.size.threshold.kB=79000
```
zakkak, line 3: We need to make these Quarkus/Mandrel specific.

My understanding is that you are increasing the threshold for Quarkus 3.6, but this will make us miss increases happening in 3.2 for instance.

zakkak (approved):
LGTM other than the threshold bumps which we should better address in another PR since it's a general issue not specific to this PR.

jerboaa (approved):
I'm fine with getting this in and do the proper threshold refactoring fix in a follow-up.

apps/quarkus-full-microprofile/threshold.properties
jerboaa: +1

Karm: @zakkak Yes. The increase is caused by OpenTelemetry being a much bigger library than OpenTracing apparently.

The "design" of the property files did not take into account the Q versions.

We can start having some per-quarkus-version property files. Perhaps per Mandrel major.minor too. 

Non-quarkus apps would have just the Mandrel major.minor?

Wondering how to make it nice and tidy without inflating the logic too much.

zakkak: > Non-quarkus apps would have just the Mandrel major.minor?

Yes

> Wondering how to make it nice and tidy without inflating the logic too much.

I need something similar in https://github.com/quarkusio/quarkus/pull/36108 and I was thinking about using a directory structure like `<quarkus-version>/<mandrel-version>/threshold.properties`. WDYT?

PR #231, opened 2023-11-29
New bike shed paint: mvn formatter:format -pl testsuite

Karm:
I set up a formatter, it reflowed everything not wrapped in  `//@formatter:off` and now I don't know whether I like it or not. Do we need it?
It's the contemporary Quarkus one + longer lines.

It does not touch the test apps, just the test code.

Perhaps we just keep the xml file for IDE?

```
mvn formatter:format -pl testsuite
```


zakkak:
> and now I don't know whether I like it or not

IMO there is no real need to reformat everything (and will only make things harder to track through git log/blame). I would prefer to have the new formatter in and reformat only edited lines as we go.

Also note that I see some changes (e.g. the removal of tests for older Mandrel versions) that don't seem to be from the reformat, and would prefer to have them in a separate commit.

Issue #237, opened 2023-12-15
Maybe add new JFR log lines to whitelist


roberttoyonaga:
### Summary
Adding this so it can be tracked and I don't forget about it. 
Depending on the outcome of https://github.com/oracle/graal/pull/8037,  new lines may need to be whitelisted for JFR.

```
[warn][jfr,setting] @Deprecated JFR events, and leak profiling are not yet supported.
[warn][jfr,setting] Exception occurred when setting value "forRemoval" for class jdk.jfr.internal.Control
Passed
```

jerboaa:
Thanks for the report @roberttoyonaga. Could you please prioritize getting those added? Thanks! We see it in CI now:
https://github.com/Karm/mandrel-integration-tests/issues/208#issuecomment-1859938636

roberttoyonaga:
> Thanks for the report @roberttoyonaga. Could you please prioritize getting those added? Thanks! We see it in CI now: [#208 (comment)](https://github.com/Karm/mandrel-integration-tests/issues/208#issuecomment-1859938636)

Yes, will do

jerboaa:
> We see it in CI now: [#208 (comment)](https://github.com/Karm/mandrel-integration-tests/issues/208#issuecomment-1859938636)

The issue we see in CI is #239 (related, but not quite the issue this is tracking). We may have to update once https://github.com/oracle/graal/pull/8037 merges upstream and close this issue.

roberttoyonaga:
Closing now because it's been resolved by https://github.com/Karm/mandrel-integration-tests/pull/242

PR #238, opened 2023-12-18
Update JFR whitelisted log lines

roberttoyonaga:
Closes: https://github.com/Karm/mandrel-integration-tests/issues/239

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
@@ -202,9 +202,18 @@ public enum WhitelistLogLines {
                         Pattern.compile(".*Exception occurred when setting value \"150/s\" for class jdk.jfr.internal.Control.*")};
             } else {
                 return new Pattern[]{
-                        // We don't support the new JFR Deprecated events annotation yet. (https://github.com/oracle/graal/pull/8037)
+                        /* We don't support the new JFR Deprecated events annotation yet.
+                         * https://github.com/oracle/graal/commit/dd8eb7436ca2ce323a74667805dd7736c6ccc972 intercepts
+                         * related calls and throws VMErrors which are handled in JDK code,
+                         * and then get logged at runtime. Allow list those log lines until they are supported.
+                         */
                         Pattern.compile(".*Exception occurred when setting value \"forRemoval\" for class jdk.jfr.internal.Control*"),
-                        // We don't support the OldObjectSample event yet. (https://github.com/oracle/graal/pull/8037)
+                        /* We don't support the OldObjectSample event yet.
+                         * https://github.com/oracle/graal/commit/dd8eb7436ca2ce323a74667805dd7736c6ccc972 intercepts
+                         * calls to adjust OldObjectSample cutoff settings and throws VMErrors which are handled in
+                         * JDK code, and then get logged at runtime. Allow list those log lines until they are
+                         * supported.
+                         */
                         Pattern.compile(".*Exception occurred when setting value \"0 ns\" for class jdk.jfr.internal.Control*"),
                         // https://github.com/oracle/graal/issues/3636
                         Pattern.compile(".*Unable to commit. Requested size [0-9]* too large.*"),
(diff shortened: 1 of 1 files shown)


zakkak (approved):
LGTM, although I would wait for 8037 to be merged before whitelisting the `@Deprecated JFR events` part.

testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
```diff
@@ -202,6 +202,10 @@ public Pattern[] get(boolean inContainer) {
                         Pattern.compile(".*Exception occurred when setting value \"150/s\" for class jdk.jfr.internal.Control.*")};
             } else {
                 return new Pattern[]{
+                        // We don't support the new JFR Deprecated events annotation yet. (https://github.com/oracle/graal/pull/8037)
+                        Pattern.compile(".*Exception occurred when setting value \"forRemoval\" for class jdk.jfr.internal.Control*"),
+                        // We don't support the OldObjectSample event yet. (https://github.com/oracle/graal/pull/8037)
+                        Pattern.compile(".*Exception occurred when setting value \"0 ns\" for class jdk.jfr.internal.Control*"),
```
jerboaa, line 208: I don't understand. Why do we see this message in CI, when https://github.com/oracle/graal/pull/8037 hasn't been merged yet? It must be coming from something else?

roberttoyonaga: https://github.com/oracle/graal/pull/8037 isn't what's creating those log lines, it just adds more text to make them clearer. [This recent GraalVM commit](https://github.com/oracle/graal/commit/dd8eb7436ca2ce323a74667805dd7736c6ccc972) is throwing the exception, which is caught in the JDK code which results in those error logs. This is the correct behaviour since we don't actually handle [those JFR features](https://github.com/openjdk/jdk/pull/16931) yet.

roberttoyonaga: The reason I mentioned waiting for https://github.com/oracle/graal/pull/8037 to be merged is so that we wouldn't have to update the whitelisted lines a second time with the modified log text.

jerboaa: OK. Can we adjust the comments then, please? I suggest something like:

```
/*
 * We don't support the new JFR Deprecated events annotation yet.
 * https://github.com/oracle/graal/commit/dd8eb7436ca2ce323a74667805dd7736c6ccc972 throws
 * VMErrors which are handled in JDK code, and then get logged at runtime. Allow list those
 * log lines until they are supported.
 */
Pattern.compile(".*Exception occurred when setting value \"forRemoval\" for class jdk.jfr.internal.Control*"),
Pattern.compile(".*Exception occurred when setting value \"0 ns\" for class jdk.jfr.internal.Control*"),
```

For the updated messages, depending on https://github.com/oracle/graal/pull/8037, we can create a new issue. Perhaps this PR should not close https://github.com/Karm/mandrel-integration-tests/issues/237 but use a separate issue explaining the situation this fixes.

Then, once 8037 merges, update the pattern lines. It should be quick to do. We need to reduce our CI noise so as to not mask actual issues we don't know of yet.

jerboaa: It's very confusing that those lines refer to a PR that is unrelated (https://github.com/oracle/graal/pull/8037). https://github.com/oracle/graal/commit/dd8eb7436ca2ce323a74667805dd7736c6ccc972 is the actual change provoking those we see in CI, right? So that should be what we refer to.

jerboaa: @roberttoyonaga I've created #239 to refer to for this PR. Thanks!

roberttoyonaga: Thanks @jerboaa, I've updated the comments.

jerboaa (approved):
Looks good.

PR #242, opened 2023-12-20
Update whitelisted JFR log lines once more

roberttoyonaga:
This is a follow up to: https://github.com/Karm/mandrel-integration-tests/pull/238 

https://github.com/oracle/graal/pull/8037#event-11298904545 Has now been merged ([internal PR](https://github.com/oracle/graal/pull/8057)), so I've adjusted the whitelisted lines. 

This should prevent things form breaking when the latest JDK and GraalVM master are used. 
This PR replaces the old whitelisted lines. I'm not sure if that's the right thing to do (in case there is a test configuration that uses a development version that isn't the latest GraalVM master). I can add the old whitelisted lines back in if that's the case.

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
@@ -202,19 +202,12 @@ public enum WhitelistLogLines {
                         Pattern.compile(".*Exception occurred when setting value \"150/s\" for class jdk.jfr.internal.Control.*")};
             } else {
                 return new Pattern[]{
-                        /* We don't support the new JFR Deprecated events annotation yet.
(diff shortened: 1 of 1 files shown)


zakkak:
> I'm not sure if that's the right thing to do (in case there is a test configuration that uses a development version that isn't the latest GraalVM master). I can add the old whitelisted lines back in if that's the case.

I think you do the right thing by removing the old ones. Since they are not part of a release I believe we would like the test to fail if they show up.

zakkak (approved):

jerboaa:
I'd like to wait with merging this until we see a failure in CI.

zakkak:
> I'd like to wait with merging this until we see a failure in CI.

@jerboaa see https://github.com/Karm/mandrel-integration-tests/issues/240

jerboaa (approved):
OK, thanks. Ship it then.

PR #255, opened 2024-04-10
Resteasy client not closed, container tests

Karm:
* Refactors JWT RestClient demo so as it cleans up dangling client even for the oldest Quarkus the app supports.
* Adds Container mode for the build perf app
* Small fixes and Whitelist reorg


Karm:
@jerboaa It's green. Not thanks to`exit 0;` :-D , but due to the fix working.

Old Quarkus needs:
```
        Client c = null;
        Response r = null;
        try {
            c = ClientBuilder.newClient();
            r = c.target("http://localhost:" + port + "/data/protected")
                    .request().header("authorization", "Bearer " + generateJWT(key))
                    .buildGet().invoke();
            return String.format("Claim value within JWT of 'custom-value' : %s", r.readEntity(String.class));
        } finally {
            if (c != null) {
                c.close();
            }
            if (r != null) {
                r.close();
            }
        }
```


New Quarkus is fine with:
```
        try (final Response response = ClientBuilder.newClient()
                .target("http://localhost:" + port + "/protected")
                .request().header("authorization", "Bearer " + generateJWT(key))
                .buildGet().invoke()) {
            return String.format("Claim value within JWT of 'custom-value' : %s", response.readEntity(String.class));
        }
```

apps/quarkus-full-microprofile/src/main/java/com/example/quarkus/secure/JWTResource.java
```diff
@@ -0,0 +1,85 @@
...
+    int port;
+
+    @GET
+    @Path("/test")
+    public String testSecureCall() {
+        if (key == null) {
+            throw new WebApplicationException("Unable to read privateKey.pem", 500);
+        }
+        // Why not try-with-resources, right?
+        // Because with certain older versions, Client is not auto-closable.
```
jerboaa, line 49: We should document which older versions still need it so that we can clean this up once we are past that.

apps/quarkus-mp-orm-dbs-awt/src/test/resources/.env
jerboaa, line 1: Is this really something we should check in? Shouldn't we set the env on the CI job instead?

Karm: @jerboaa It should not be needed and it is not for a modern Quarkus as it manages to inject those via container props. The old, as in Quarkus 2, for some reason fails to do that, so the quarkus app running in-container testing fails to log in to the database. The solution I came up with is to copy the .env file to the runtime image so as the app loads it.  It works.

Additional thing is that we have to copy something if there are no .so files with older Mnadrel, this file, alwasys present, serves that purpose too.

apps/quarkus-full-microprofile/src/main/java/com/example/quarkus/secure/JWTResource.java
Karm: Will add comment.

PR #258, opened 2024-05-07
Scripts demo Runnerid info

Karm:
Instead of drawing a picture, I copy-pasted bits and pieces from Jenkins/GHA/local run scripts into an aux directory with the goal to sketch out how I think we should upload the stats with Runner info.

When we run with Quarkus Quickstarts, we could use the very same logic as with Quarkus Integration Tests, i.e. one Runner Info, many Image Build Stats.


aux_scripts/q_common_modules.sh
```diff
@@ -0,0 +1,35 @@
...
+"2.16.12.Final" \
+"3.2.12.Final" \
+"3.8.4" \
+"3.9.5" \
+"3.10.0" \
+"main"
+)
+
+for QTAG in "${QTAGS[@]}"; do
+  curl https://raw.githubusercontent.com/quarkusio/quarkus/main/.github/native-tests.json \
```
zakkak, line 19: Shouldn't this fetch the native-tests.json file of the corresponding `QTAG` instead of `main`?

apps/quarkus-mp-orm-dbs-awt/src/main/resources/application.properties
```diff
@@ -17,7 +17,7 @@ quarkus.datasource.db2.username=quarkus
 quarkus.datasource.db2.password=quarkus
 quarkus.datasource.db2.devservices.enabled=true
 quarkus.datasource.db2.devservices.port=49157
-quarkus.datasource.db2.devservices.image-name=docker.io/library/mariadb:10.11
+quarkus.datasource.db2.devservices.image-name=quay.io/mariadb-foundation/mariadb-devel:10.11
```
zakkak, line 20: I guess that's needed for debugging purposes? Aren't the regular logs enough?

zakkak (changes requested):
LGTM, other than the script using the native tests from `main` instead of the corresponding Quarkus tag.

I wonder however whether it's worth keeping these demos in this repository. They look more related to the `collector` repository than the `mandrel-integration-tests` repository.

apps/quarkus-mp-orm-dbs-awt/src/main/resources/application.properties
Karm: @zakkak As a matter of fact it is due to DockerHub rate limiting and this being a quay.io mariadb from a publisher I decided to trust...

aux_scripts/q_common_modules.sh
Karm: :scream:  Yes. Thx!

Karm:
> LGTM, other than the script using the native tests from `main` instead of the corresponding Quarkus tag.
> 
> I wonder however whether it's worth keeping these demos in this repository. They look more related to the `collector` repository than the `mandrel-integration-tests` repository.

Hmm, I don't know. Feels like it could be in both. I'll add it here as the Integration Tests use it and I look into those to configure CI uploads whereas Collector has its own tests as demos...    :man_shrugging:

Issue #261, opened 2024-06-26
Allow list DynamicProxyConfigurationResources related warning so as to reduce test noise


jerboaa:
In latest Mandrel for JDK 23 and Mandrel for JDK 24 builds we see these warnings printed for mandrel-integration-tests:

```
 Error:    DebugSymbolsTest.debugSymbolsQuarkus:306 build-and-run.log log should not contain error or warning lines that are not whitelisted. See /home/runner/work/mandrel/mandrel/mandrel-integration-tests/testsuite/target/archived-logs/org.graalvm.tests.integration.DebugSymbolsTest/debugSymbolsQuarkus/build-and-run.log and check these offending lines: 
Warning: Option 'DynamicProxyConfigurationResources' is deprecated and might be removed in a future release. Please refer to the GraalVM release notes. ==> expected: <true> but was: <false>
```

The actual issue got reported upstream here:
https://github.com/oracle/graal/issues/9057

Until the upstream issue gets fixed we ought to allow-list that warning so as to reduce test noise.

jerboaa:
Steps to reproduce (for example):

```
export JAVA_HOME=/path/to/mandrel-build-based-on-jdk23
export GRAALVM_HOME=$JAVA_HOME
export PATH=$JAVA_HOME/bin:$PATH
mvn clean verify -Dquarkus.version=999-SNAPSHOT -Ptestsuite -Dtest=RuntimesSmokeTest#quarkusFullMicroProfile
```

zakkak:
> Until the upstream issue gets fixed we ought to allow-list that warning so as to reduce test noise.

Note that this is now "fixed" in upstream by https://github.com/oracle/graal/commit/5fc14c42fd8bbad0c8e661b4ebd8f96255f86e6b which essentially improves the warning message but doesn't remove it.

We need to either patch the tests to use the new metadata file with the latest version of GraalVM or whitelist the warning.

jerboaa:
> We need to either patch the tests to use the new metadata file with the latest version of GraalVM or whitelist the warning.

OK. I'm leaning towards allow-listing the warning.

zakkak:
On second thought we can't really fix this here without any whitelisting due to https://github.com/quarkusio/quarkus/issues/41016, so +1 on just whitelisting for the time being.

jerboaa:
Re-opening. We need to allow-list the warning for the JFR test, `JFRTest.jfrPerfTest`, as well:

```
[...]

PR #264, opened 2024-06-26
Allow lists for changed metadata config files

jerboaa:
Note that I've also added another allow-list entry that my local maven spits out. I couldn't get the tests to pass otherwise. Hope that's OK.

Closes: #261


zakkak:
> Note that I've also added another allow-list entry that my local maven spits out. I couldn't get the tests to pass otherwise. Hope that's OK.

If this is happening only locally I wouldn't include it, but I will let @Karm make the call.
You might want to check if something is setting `MAVEN_OPTS` in your environment.

jerboaa:
> > Note that I've also added another allow-list entry that my local maven spits out. I couldn't get the tests to pass otherwise. Hope that's OK.
> 
> If this is happening only locally I wouldn't include it, but I will let @Karm make the call. You might want to check if something is setting `MAVEN_OPTS` in your environment.

@zakkak Apparently it's a maven 3.9.1 thing. The upstream maven change seems to be this: https://issues.apache.org/jira/browse/MNG-7706 That has effects on maven-surefire and other maven plugins. So depending on the version we pull in (from quarkus?) the warning will be there. I don't think this is something we can fix. I'll mention this issue in the allow list entry.

zakkak (approved):

Issue #282, opened 2024-09-23
New serialization tests fail on Mandrel for JDK 21 (21.0.5-ea), 24.2


jerboaa:
The tests added with #279 now fail in weekly CI. Not sure if this is expected. They fail with:

```
2024-09-21T01:08:31.7516407Z [ERROR] Tests run: 15, Failures: 1, Errors: 0, Skipped: 6, Time elapsed: 524.743 s <<< FAILURE! - in org.graalvm.tests.integration.AppReproducersTest
2024-09-21T01:08:31.7518580Z [ERROR] forSerialization21Test{TestInfo}  Time elapsed: 57.868 s  <<< FAILURE!
2024-09-21T01:28:35.0237823Z [ERROR] Failures: 
2024-09-21T01:28:35.0242514Z [ERROR]   AppReproducersTest.forSerialization21Test:830->forSerialization:879 The output of the HotSpot and native-image runs must be the same. ==> expected: <0 0 [Z [Z true
2024-09-21T01:28:35.0346143Z [ERROR] Tests run: 22, Failures: 1, Errors: 0, Skipped: 6
2024-09-21T01:28:35.0349144Z [INFO] testsuite .......................................... FAILURE [28:54 min]
2024-09-21T01:28:35.0361174Z [ERROR]   mvn <args> -rf :testsuite
```

Looks to be related to https://github.com/oracle/graal/issues/9581

See: https://github.com/graalvm/mandrel/actions/runs/10968086869/job/30459449251#step:11:21167

jerboaa:
@Karm Is this supposed to fail?

Karm:
@jerboaa Hmm, it is supposed to fail as in: yes, it is not fixed. Although it should have been excluded from the execution.

jerboaa:
Right, thanks!

Karm:
> Why so? Is it to reduce noise on the CI?

Yes.

> AFAIK we want this tested with 23.1

Yes.

Although I have no idea how long it might take. I will ask Vojin for pointers and might try to take a look.

jerboaa:
> > Although it should have been excluded from the execution.
> 
> Why so? Is it to reduce noise on the CI?

Please exclude the test until a fix is ready. It's not practical to add tests which are known to fail for now. Triage is painful enough as it is.

jerboaa:
Well it turns out the test also fails on latest graal master (JDK 24-based):

```
Error:  Failures: 
[...]

PR #286, opened 2024-09-26
Disable String[] check in serialization test

zakkak:
Works around https://github.com/oracle/graal/issues/9581 to [avoid CI noise](https://github.com/Karm/mandrel-integration-tests/issues/282#issuecomment-2376449083)


jerboaa (approved):
Thanks!

Issue #285, opened 2024-09-26
AppReproducersTest#timezonesBakedIn fails with latest GraalVM for JDK 24


jerboaa:
This is similar to https://github.com/quarkusio/quarkus/issues/43436 caused by a recent GraalVM change (https://github.com/oracle/graal/pull/9694). Since the default language is no longer being set in stone at build time, the runtime language needs to be explicitly set to the desired one for Mandrel for JDK 24.

The test failure looks like this:

```
[...]
2024-09-26 11:22:03.665 INFO  [o.g.t.i.u.Commands$ProcessRunner] (run) Command: [native-image, -J-Duser.country=CA, -J-Duser.language=fr, -jar, target/timezones.jar, target/timezones]
========================================================================================================================
GraalVM Native Image: Generating 'timezones' (executable)...
========================================================================================================================
For detailed information and explanations on the build output, visit:
https://github.com/oracle/graal/blob/master/docs/reference-manual/native-image/BuildOutput.md
------------------------------------------------------------------------------------------------------------------------
[1/8] Initializing...                                                                                    (3,3s @ 0,07GB)
 Java version: 24-beta+15-ea, vendor version: Mandrel-24.2.0-dev
 Graal compiler: optimization level: 2, target machine: x86-64-v3
 C compiler: gcc (redhat, x86_64, 13.3.1)
 Garbage collector: Serial GC (max heap size: 80% of RAM)
 1 user-specific feature(s):
 - com.oracle.svm.thirdparty.gson.GsonFeature
------------------------------------------------------------------------------------------------------------------------
Build resources:
 - 26,49GB of memory (42,4% of 62,55GB system memory, determined at start)
 - 12 thread(s) (100,0% of 12 available processor(s), determined at start)
[2/8] Performing analysis...  [****]                                                                     (6,6s @ 0,37GB)
    3 334 reachable types   (69,9% of    4 767 total)
    3 837 reachable fields  (42,9% of    8 943 total)
   15 563 reachable methods (44,0% of   35 351 total)
    1 033 types,     7 fields, and   141 methods registered for reflection
       57 types,    57 fields, and    52 methods registered for JNI access
        4 native libraries: dl, pthread, rt, z
[3/8] Building universe...                                                                               (1,2s @ 0,37GB)
[4/8] Parsing methods...      [*]                                                                        (0,7s @ 0,38GB)
[5/8] Inlining methods...     [***]                                                                      (0,6s @ 0,39GB)
[...]

PR #287, opened 2024-09-26
Fix AppReproducersTest#timezonesBakedIn

jerboaa:
With GraalVM prior to `24.2.0` the default locale (and included locale) for a native image was determined by the -Duser.language and -Duser.country settings as image build time. With GraalVM `>= 24.2.0` the build settings of language/country have no effect in terms of included locales in the resulting native image. Therefore, `-H:IncludeLocales`, needs to be used at build time instead. Also at image runtime the default language is determined by the system locale. For the purpose of the test we'd want the French language which we need to set at image runtime when the test runs.

Tested with `GraalVM community 23+37` (which uses the old scheme) and with a latest JDK 24-based mandrel build which uses the new scheme. Both pass.

24.1:
```
[...]
Command: native-image -J-Duser.country=CA -J-Duser.language=fr -jar target/timezones.jar target/timezones
[...]
Command: ./target/timezones
jeu. sept. 26 12:48:51 CEST 2024
heure normale de l’Europe centrale
```

24.2
```
[...]
Command: native-image -H:IncludeLocales=fr-CA -jar target/timezones.jar target/timezones
[...]
Command: ./target/timezones -Duser.language=fr
jeu. sept. 26 12:53:57 CEST 2024
heure normale d’Europe centrale
```

Closes: #285

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
@@ -85,8 +85,15 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 public class AppReproducersTest {
 
     private static final Logger LOGGER = Logger.getLogger(AppReproducersTest.class.getName());
+    private static final String LOCALEINCLUDES_SWITCH_REPLACEMENT_1_MANDREL_PRE_24_2_0 = "-J-Duser.country=CA";
+    private static final String LOCALEINCLUDES_SWITCH_REPLACEMENT_2_MANDREL_PRE_24_2_0 = "-J-Duser.language=fr";
+    private static final String LOCALEINCLUDES_SWITCH_REPLACEMENT_1_MANDREL_POST_24_2_0 = "-H:IncludeLocales=fr-CA";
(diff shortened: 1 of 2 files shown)


zakkak (approved):
LGTM. I am just wondering whether we should instead fix this on the Quarkus side by adding `user.language`-`user.country` to the included locales like we do for `quarkus.default-locale` and `quarkus.native.user-language`-`quarkus.native.country` in https://github.com/quarkusio/quarkus/pull/43448

jerboaa:
> LGTM.

Thanks for the review.

> I am just wondering whether we should instead fix this on the Quarkus side by adding `user.language`-`user.country` to the included locales like we do for `quarkus.default-locale` and `quarkus.native.user-language`-`quarkus.native.country` in [quarkusio/quarkus#43448](https://github.com/quarkusio/quarkus/pull/43448)

I'm not sure we should. Having the build system locale included by default isn't something which I think all users are aware of. Maybe raise as an issue and see what the desired outcome of it is.

zakkak:
> Maybe raise as an issue and see what the desired outcome of it is.

Done in https://github.com/quarkusio/quarkus/discussions/43533

jerboaa:
@Karm OK with this? CI looks good too :)

Karm (approved):
THX @jerboaa

Issue #293, opened 2024-10-24
Mandrel integration tests suite fails to parse GraalVM/Mandrel version when warnings are present in the output


zakkak:
E.g. it throws

```
2024-10-24 11:51:23.628 WARN  [o.g.t.i.u.v.UsedVersion$MVersion] (of) Failed to correctly parse native-image version command output. Is it on PATH? Unknown version format? Output reads in 4 lines, see them in an array: [WARNING: Unknown module: org.graalvm.truffle specified to --enable-native-access, native-image 24-beta 2025-03-18, OpenJDK Runtime Environment Mandrel-24.2.0-deve05805b1084 (build 24-beta+20-ea), OpenJDK 64-Bit Server VM Mandrel-24.2.0-deve05805b1084 (build 24-beta+20-ea, mixed mode)]
```

when the output is:

```
WARNING: Unknown module: org.graalvm.truffle specified to --enable-native-access
native-image 24-beta 2025-03-18
OpenJDK Runtime Environment Mandrel-24.2.0-deve05805b1084 (build 24-beta+20-ea)
OpenJDK 64-Bit Server VM Mandrel-24.2.0-deve05805b1084 (build 24-beta+20-ea, mixed mode)
```

We should ignore the warning and still parse the version to avoid test failures. If we want to test for warnings being present in the `version` command, then we should have a separate test for it.

jerboaa:
+1

PR #294, opened 2024-10-24
Ignore irrelevant output when parsing Mandrel version

zakkak:
Drop everything until we see "GraalVM" or "native-image" in the output.
This way we can ignore warnings and other output that might be present.

Fixes https://github.com/Karm/mandrel-integration-tests/issues/293

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/versions/UsedVersion.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/versions/UsedVersion.java
@@ -24,6 +24,7 @@ import org.graalvm.tests.integration.utils.Commands;
 import org.jboss.logging.Logger;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


jerboaa (commented):
The quarkus version of this is:

```
            List<String> lines = stringOutput.lines()
                    .dropWhile(l -> !l.startsWith("GraalVM") && !l.startsWith("native-image"))
                    .toList();
```

Should we use the same?

zakkak:
I avoided this to keep it [JDK 11 compatible](https://github.com/Karm/mandrel-integration-tests/blob/62113f234db68984b33e1ba7a161cfbd7ffcbcc6/pom.xml#L38-L39). Not sure if that's still necessary though. I am OK with either implementation as long as @Karm is happy with it :)

jerboaa:
> I avoided this to keep it [JDK 11 compatible](https://github.com/Karm/mandrel-integration-tests/blob/62113f234db68984b33e1ba7a161cfbd7ffcbcc6/pom.xml#L38-L39). Not sure if that's still necessary though. I am OK with either implementation as long as @Karm is happy with it :)

Ugh. We should clean this up at some point. Those old versions don't make sense to me anymore to lose any sleep over it.

jerboaa (approved):

PR #296, opened 2024-11-19
Race condition with RHEL9 Podman version, Whitelists, Thresholds

Karm:
There is a combination of Podman version and Java version where this piece of code hangs indefinitely:

```java
final URLConnection c = new URL(url).openConnection();
c.setRequestProperty("Accept", "*/*");
c.setConnectTimeout(500);
try (Scanner scanner = new Scanner(c.getInputStream(), StandardCharsets.UTF_8)) {
                                    ^^^^^^^^^^^^^^^^^
   scanner.useDelimiter("\\A");
   webPage = scanner.hasNext() ? scanner.next() : "";
}
```

on Scanner trying to read the input stream.  It seems that what used to be rather simple:

    a) an exception is thrown, because Quarkus is not yet ready
    b) a request response is populated in the buffer and read

is now more convoluted with some third state 

    c) no exception is thrown, but the buffer is not populated, Scanner hangs

I included a check for word `started` in the container log and that seems to alleviate this situation.

### DIFF
--- a/apps/quarkus-spöklik-encoding/threshold.conf
+++ b/apps/quarkus-spöklik-encoding/threshold.conf
@@ -20,4 +20,7 @@ linux.container.time.to.first.ok.request.threshold.ms=1500
 linux.container.RSS.threshold.kB=55000
 @IfQuarkusVersion(min ="3.7.0")
 linux.container.executable.size.threshold.kB=51313
-
+@IfQuarkusVersion(min ="3.8.6")
+linux.container.executable.size.threshold.kB=51991
+@IfQuarkusVersion(min ="3.15.1")
+linux.container.executable.size.threshold.kB=52860
--- a/testsuite/src/it/java/org/graalvm/tests/integration/RuntimesSmokeTest.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/RuntimesSmokeTest.java
@@ -42,6 +42,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
 import static org.graalvm.tests.integration.utils.Commands.QUARKUS_VERSION;
@@ -59,6 +60,7 @@ import static org.graalvm.tests.integration.utils.Commands.removeContainer;
 import static org.graalvm.tests.integration.utils.Commands.runCommand;
 import static org.graalvm.tests.integration.utils.Commands.stopAllRunningContainers;
 import static org.graalvm.tests.integration.utils.Commands.stopRunningContainer;
+import static org.graalvm.tests.integration.utils.Commands.waitForContainerLogToMatch;
 import static org.graalvm.tests.integration.utils.Commands.waitForTcpClosed;
 
 /**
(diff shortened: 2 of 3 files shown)


Karm:
Internal CI is happy too. No more hangs across crun/podman versions.

testsuite/src/it/java/org/graalvm/tests/integration/RuntimesSmokeTest.java
```diff
@@ -106,6 +108,10 @@ public void testRuntime(TestInfo testInfo, Apps app, Map<String, String> switchR
             Logs.appendln(report, appDir.getAbsolutePath());
             Logs.appendlnSection(report, String.join(" ", cmd));
 
+            if (app.runtimeContainer != ContainerNames.NONE) {
+                waitForContainerLogToMatch(app.runtimeContainer.name, Pattern.compile(".*started.*"), 5, 1, TimeUnit.SECONDS);
```
zakkak, line 112: That makes sense. Do you think lowering the timeout and sleep time would be problematic? Waiting 5 seconds to timeout seems quite long to me, as well as waiting a whole second to retry.

apps/quarkus-spöklik-encoding/threshold.conf
```diff
@@ -20,4 +20,7 @@ linux.container.time.to.first.ok.request.threshold.ms=1500
 linux.container.RSS.threshold.kB=55000
 @IfQuarkusVersion(min ="3.7.0")
 linux.container.executable.size.threshold.kB=51313
-
+@IfQuarkusVersion(min ="3.8.6")
+linux.container.executable.size.threshold.kB=51991
+@IfQuarkusVersion(min ="3.15.1")
+linux.container.executable.size.threshold.kB=52860
```
zakkak, line 26: How are these related to the issue at hand?

testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
```diff
@@ -65,8 +65,19 @@ public Pattern[] get(boolean inContainer) {
...
                     Pattern.compile(".*Please ignore if you don't have images in an authenticated registry.*"),
+                    // Common new Q versions
+                    Pattern.compile(".*io.quarkus.narayana.jta.runtime.graal.DisableLoggingFeature.*"),
+                    // Podman / Docker extension incompatibilities with Podman versions
+                    Pattern.compile(".*Database JDBC URL \\[undefined/unknown\\].*"),
+                    Pattern.compile(".*Database driver: undefined/unknown.*"),
+                    Pattern.compile(".*Autocommit mode: undefined/unknown.*"),
+                    Pattern.compile(".*Minimum pool size: undefined/unknown.*"),
+                    Pattern.compile(".*Isolation level: <unknown>.*"),
+                    Pattern.compile(".*Maximum pool size: undefined/unknown.*"),
```
zakkak, line 80: It feels like you are doing more things than advertised in this PR. Could you please update the PR title/description to reflect these changes as well?

zakkak (changes requested):

apps/quarkus-spöklik-encoding/threshold.conf
Karm: Totally unrelated, I was just stabilizing the run as I was testing it. Gonna amend the PR to note it.

zakkak (approved):

PR #319, opened 2025-03-26
Changes default container runtime to podman

Karm:
Historically, this TS assumed docker-ce the default container runtime. That has changed over the years to a point where we have either Podman or Podman Desktop installed everywhere and we keep cluttering test envs with an explicit switch to podman.
A notable exception is GHA.

I think we can switch the default to podman now...

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/Commands.java
@@ -78,7 +78,7 @@ public class Commands {
 
     private static final Logger LOGGER = Logger.getLogger(Commands.class.getName());
 
-    public static final String CONTAINER_RUNTIME = getProperty("QUARKUS_NATIVE_CONTAINER-RUNTIME", "docker");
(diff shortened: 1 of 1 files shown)


jerboaa (approved):

Karm:
:red_circle: :x: 

Huh... 

Related to https://github.com/quarkusio/quarkus-images and purging of old images...?

zakkak:
> A notable exception is GHA.

It looks like [linux github runners include podman as well](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md), the reason we use docker is for increased coverage? Legacy? Or some other limitation?

Karm:
> > A notable exception is GHA.
> 
> It looks like [linux github runners include podman as well](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md), the reason we use docker is for increased coverage? Legacy? Or some other limitation?

Dropping `docker` from GHA then :+1:

zakkak (approved):
LGTM, thanks

Karm:
Fails unrelated to this PR.

PR #331, opened 2025-05-11
"devel-latest" is UBI9

Karm:
Related to https://github.com/quarkusio/quarkus-images/pull/299


Karm:
This should not be needed any more. I fixed the bug where the "ubi9" suffix was not visible to the testsuite.

PR #335, opened 2025-05-20
Tests runtime and sys props from virtual threads

Karm:
Tests issue related to: https://github.com/oracle/graal/pull/10109/files as a part of backports testing.

### DIFF
--- a/apps/vthread_props/src/main/java/vthread_props/Main.java
+++ b/apps/vthread_props/src/main/java/vthread_props/Main.java
@@ -31,6 +31,9 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 public class Main {
+
+    private static long TIMEOUT_MS = 100;
+
     public static void main(String[] args) {
         final String r = String.format("%s %s %s %s %s %s",
                 runTest("System Properties Access", Main::testSystemPropertiesAccess),
@@ -58,7 +61,7 @@ public class Main {
                 e.printStackTrace();
             }
         });
-        if (!vt.join(Duration.ofMillis(10))) {
+        if (!vt.join(Duration.ofMillis(TIMEOUT_MS))) {
             System.out.println("Error: Virtual thread did not join in time");
             return false;
         }
@@ -84,7 +87,7 @@ public class Main {
                 e.printStackTrace();
             }
         });
-        if (!vt.join(Duration.ofMillis(20))) {
+        if (!vt.join(Duration.ofMillis(TIMEOUT_MS))) {
             System.out.println("Error: Virtual thread did not join in time");
             return false;
         }
@@ -110,7 +113,7 @@ public class Main {
                 e.printStackTrace();
             }
         });
-        if (!vt.join(Duration.ofMillis(15))) {
+        if (!vt.join(Duration.ofMillis(TIMEOUT_MS))) {
             System.out.println("Error: Virtual thread did not join in time");
             return false;
         }
@@ -139,7 +142,7 @@ public class Main {
                     }
                 });
             }
-            if (!l.await(15, TimeUnit.MILLISECONDS)) {
+            if (!l.await(TIMEOUT_MS, TimeUnit.MILLISECONDS)) {
                 System.out.println("Error: Virtual threads did not join in time.");
                 return false;
             }
@@ -169,7 +172,7 @@ public class Main {
                 e.printStackTrace();
             }
         });
-        if (!vt.join(Duration.ofMillis(5))) {
+        if (!vt.join(Duration.ofMillis(TIMEOUT_MS))) {
             System.out.println("Error: Virtual thread did not join in time");
             return false;
         }
@@ -202,7 +205,7 @@ public class Main {
                 e.printStackTrace();
             }
         });
-        if (!vt.join(Duration.ofMillis(10))) {
+        if (!vt.join(Duration.ofMillis(TIMEOUT_MS))) {
             System.out.println("Error: Virtual thread did not join in time");
             return false;
         }
--- a/apps/vthread_props/src/main/java/vthread_props/Main.java
+++ b/apps/vthread_props/src/main/java/vthread_props/Main.java
@@ -1,0 +1,223 @@
+/*
+ * Copyright (c) 2025, Red Hat Inc. All rights reserved.
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package vthread_props;
+
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.Properties;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class Main {
+    public static void main(String[] args) {
+        final String r = String.format("%s %s %s %s %s %s",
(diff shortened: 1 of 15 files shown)


apps/vthread_props/src/main/java/vthread_props/Main.java
```diff
@@ -0,0 +1,223 @@
...
+        System.out.println("\n=== RESULT: " + r + " ===");
+    }
+
+    private static boolean testSystemPropertiesAccess() throws Exception {
+        final CountDownLatch l = new CountDownLatch(1);
+        final AtomicReference<Properties> p = new AtomicReference<>();
+        final AtomicReference<Exception> exception = new AtomicReference<>();
+        final Thread vt = Thread.ofVirtual().start(() -> {
+            try {
+                System.out.println("Virtual: " + (Thread.currentThread().isVirtual() ? "True" : "ERROR"));
```
zakkak, line 50: The `True vs ERROR` printing here is a bit strange. I assume you do it that way so that log parsing will catch the "error" is that right?

If so I would prefer an `if-statement` controlling for that case and printing the Error if `isVirtual` is `false`. That would remove any potential ambiguity about what's causing this "ERROR".

apps/vthread_props/src/main/java/vthread_props/Main.java
```diff
@@ -0,0 +1,223 @@
...
+                runTest("Multiple Time Operations", Main::testMultipleTimeOperations),
+                runTest("Virtual Thread Properties", Main::testVirtualThreadProperties),
+                runTest("Process and System Operations", Main::testProcessAndSystemOperations));
+        System.out.println("\n=== RESULT: " + r + " ===");
+    }
+
+    private static boolean testSystemPropertiesAccess() throws Exception {
+        final CountDownLatch l = new CountDownLatch(1);
+        final AtomicReference<Properties> p = new AtomicReference<>();
+        final AtomicReference<Exception> exception = new AtomicReference<>();
```
zakkak, line 47: I like the use of atomics instead of `volatile` :)

apps/vthread_props/src/main/java/vthread_props/Main.java
```diff
@@ -0,0 +1,223 @@
...
+                System.out.println("Accessed " + props.size() + " system properties");
+            } catch (Exception e) {
+                exception.set(e);
+                e.printStackTrace();
+            } finally {
+                l.countDown();
+            }
+        });
+        l.await(1, TimeUnit.MICROSECONDS);
+        vt.join();
```
zakkak, line 62: Why do you need both? Wouldn't only `vt.join()` be enough, removing the need for the countdown latch?

apps/vthread_props/src/main/java/vthread_props/Main.java
```diff
@@ -0,0 +1,223 @@
...
+                        System.out.println("Thread " + threadNum + ": " + LocalDateTime.now() + " in zone " + ZoneId.systemDefault());
+                    } catch (Exception e) {
+                        exception.set(e);
+                        e.printStackTrace();
+                    } finally {
+                        l.countDown();
+                    }
+                });
+            }
+            l.await(1, TimeUnit.MICROSECONDS);
```
zakkak, line 138: Not checking `await`'s return value and setting the timeout to 1 microseconds seems a bit "risky". I would expect the test to fail in case of a timeout.

testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
```diff
@@ -924,6 +925,113 @@ public void jdkReflections(TestInfo testInfo, Apps app) throws IOException, Inte
...
+    @IfMandrelVersion(min = "23.1.8", max = "23.1.999", inContainer = true)
+    public void vthreadsPropsContainer23_1Test(TestInfo testInfo) throws IOException, InterruptedException {
+        vthreadsProps(testInfo, Apps.VTHREADS_PROPS_BUILDER_IMAGE);
+    }
+
+    @Test
+    @IfMandrelVersion(min = "23.1.8", max = "23.1.999")
+    public void vthreadsProps23_1Test(TestInfo testInfo) throws IOException, InterruptedException {
+        vthreadsProps(testInfo, Apps.VTHREADS_PROPS);
+    }
```
zakkak, line 939: The fix is not yet backported to 23.1, I would prefer to enable this only after the fact.

apps/vthread_props/src/main/java/vthread_props/Main.java
Karm: Thanks, this is me playing with it and leaving it half-there. Fixing...

apps/vthread_props/src/main/java/vthread_props/Main.java
Karm: I was playing wit how long it takes and left it half way through. Fixing...

apps/vthread_props/src/main/java/vthread_props/Main.java
Karm: Ack.

testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
Karm: It's mentioned here: graalvm-community-jdk21u/issues/66 and I was about to do the backport. Wrote a test first.

Karm:
Ad UBI10: It keeps being flaky in CI due to issues with the registry and with microdnf instabilities. Beyond our control. I am considering disabling UBI10 until resolved. We do have *CentOS 10* though, so any glibc runtime situations should be covered by that...I assume.

Karm:
```
 > [2/4] RUN microdnf install freetype fontconfig -y:
0.065 
0.065 (microdnf:1): librhsm-WARNING **: 12:29:19.639: Found 0 entitlement certificates
0.069 
0.069 (microdnf:1): librhsm-WARNING **: 12:29:19.643: Found 0 entitlement certificates
0.100 Downloading metadata...
0.215 Downloading metadata...
0.351 Downloading metadata...
0.375 error: Parsing armored OpenPGP packet(s) failed
------
Dockerfile.ubi10:2
--------------------
   1 |     FROM registry.access.redhat.com/ubi10/ubi-minimal:10.0
   2 | >>> RUN microdnf install freetype fontconfig -y
   3 |     WORKDIR /work/
   4 |     RUN chmod "g+rwX" /work \
--------------------
ERROR: failed to solve: process "/bin/sh -c microdnf install freetype fontconfig -y" did not complete successfully: exit code: 1
```

I disabled UBI10 in [12b1031](https://github.com/Karm/mandrel-integration-tests/pull/335/commits/12b1031881ad1688001113fab10f96fdffcc5b7c) until further notice...

testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
zakkak: I know, I am just saying that I would like to have the test enabled only after doing the backport :)

I suggest commenting it out for the time being.

apps/vthread_props/src/main/java/vthread_props/Main.java
```diff
@@ -0,0 +1,232 @@
...
+                }
+                final Properties props = System.getProperties();
+                p.set(props);
+                System.out.println("Accessed " + props.size() + " system properties");
+            } catch (Exception e) {
+                exception.set(e);
+                e.printStackTrace();
+            }
+        });
+        if (!vt.join(Duration.ofMillis(10))) {
```
zakkak, line 61: This is expected to rarely fail. Pushing the timeout limit feels of low value since we are not testing the speed of virtual threads here. I would prefer a longer timeout (possibly one second) even if that translates to waiting a bit more in case of failures.

zakkak (changes requested):

testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
Karm: @zakkak Commented out in [00ba907](https://github.com/Karm/mandrel-integration-tests/pull/335/commits/00ba907c0ad4e008a1aa7d9501b024d098bc0dec)

apps/vthread_props/src/main/java/vthread_props/Main.java
Karm: My intention is not to measure speed, rather to catch if anything weird starts to happen, like if an operation that used to take 5ms suddenly takes more than 3x that.

Yet it felt like overengineering it to use thresholds.properties.

I bumped it an order of magnitude to 100ms in [e4d6e0e](https://github.com/Karm/mandrel-integration-tests/pull/335/commits/e4d6e0ec78245dda84c665a128dc50497ddc33fa).

I propose that if such timeout is reached, there is an issue to investigate.  It passes on our weakest systems: GHA runners.

zakkak (approved):

Issue #338, opened 2025-06-03
apps/vthread_props doesn't compile with JDK 17


jerboaa:
`apps/vthread_props` is testing virtual threads which became GA in JDK 21 and JDK 17 don't support them. Therefore, this application fails to compile with a JDK lower than 21:

```
$ JAVA_HOME=/usr/lib/jvm/java-17-openjdk mvn clean package
[INFO] Scanning for projects...
[INFO] 
[INFO] --------------------< vthread-props:vthread_props >---------------------
[INFO] Building vthread_props 1
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- clean:3.2.0:clean (default-clean) @ vthread_props ---
[INFO] Deleting /disk/graal/upstream-sources/mandrel-integration-tests/apps/vthread_props/target
[INFO] 
[INFO] --- resources:3.3.1:resources (default-resources) @ vthread_props ---
[INFO] skip non existing resourceDirectory /disk/graal/upstream-sources/mandrel-integration-tests/apps/vthread_props/src/main/resources
[INFO] 
[INFO] --- compiler:3.11.0:compile (default-compile) @ vthread_props ---
[INFO] Changes detected - recompiling the module! :source
[INFO] Compiling 1 source file with javac [debug target 17] to target/classes
[INFO] Some messages have been simplified; recompile with -Xdiags:verbose to get full output
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /disk/graal/upstream-sources/mandrel-integration-tests/apps/vthread_props/src/main/java/vthread_props/Main.java:[51,33] cannot find symbol
  symbol:   method ofVirtual()
  location: class java.lang.Thread
[ERROR] /disk/graal/upstream-sources/mandrel-integration-tests/apps/vthread_props/src/main/java/vthread_props/Main.java:[53,44] cannot find symbol
  symbol:   method isVirtual()
  location: class java.lang.Thread
[ERROR] /disk/graal/upstream-sources/mandrel-integration-tests/apps/vthread_props/src/main/java/vthread_props/Main.java:[64,39] incompatible types: java.time.Duration cannot be converted to long
[ERROR] /disk/graal/upstream-sources/mandrel-integration-tests/apps/vthread_props/src/main/java/vthread_props/Main.java:[77,33] cannot find symbol
[...]

PR #340, opened 2025-06-05
Fix vthreads app test when JAVA_HOME points to JDK 17

jerboaa:
In some scenarios the Java build is using JDK 17 (default for quarkus), but the vthreads integration test doesn't work on a Java version lower than 21. Since these tests are native integration tests and the test executes on Mandrel 24.2+ (JDK 24+) only so far, we can be sure that the javac from the GraalVM/Mandrel installation supports virtual threads. Use that over the one set in JAVA_HOME. Derive the value from where native-image is located.

The container case is being handled [elsewhere](https://github.com/Karm/mandrel-integration-tests/blob/6a5e500c0999e289c71ac0e1ce5b9c1f5e8dd407/testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java#L548-L551).

Closes: #338

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
@@ -55,12 +55,10 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.awt.image.BufferedImage;
-import java.io.BufferedInputStream;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


Karm:
@jerboaa Hmm, the case where JAVA_HOME is 17 the case of builder images, i.e. you have JDK 17 locally and then you use JDK 21 builder image. To test this, the builder image flow uses the javac from the image on purpose:

https://github.com/Karm/mandrel-integration-tests/blob/master/testsuite/src/it/java/org/graalvm/tests/integration/utils/BuildAndRunCmds.java#L550

The test is deliberately **disabled** for Mandrel 23.0 JDK 17...    

Are you saying that we should enable a case where we run it hosted, without builder image, and the javac is from Java 17 and the native-image is from Java 21?

jerboaa:
> Are you saying that we should enable a case where we run it hosted, without builder image, and the javac is from Java 17 and the native-image is from Java 21?

@Karm Yes. That's the setup we use in our main CI for Quarkus. The base JDK for quarkus is still 17. See the linked issue which has Mandrel IT failures due to that.

jerboaa:
GHA failure seems unrelated to this.

zakkak (approved):
LGTM, thanks

jerboaa:
@Karm OK with this too?

jerboaa:
Recent CI failures because of this issue are here:
Windows: https://github.com/Karm/mandrel-integration-tests/issues/301#issuecomment-2947829072
Linux: https://github.com/Karm/mandrel-integration-tests/issues/300#issuecomment-2947828904

Karm (approved):
THX

@jerboaa O.K., thx.

I did not hit this, because I use JDK 17 to build Quarkus, but then JDK 21 to actually run it in the scripts.

jerboaa:
Thank you!

Issue #369, opened 2025-09-26
Adjust extra warning log lines matching for graal/master updates


jerboaa:
Graal master branch recently switched to be JDK 25-based again. Specifically Labs OpenJDK 25, where versions look like this:

```
native-image 25 2025-09-16
GraalVM Runtime Environment GraalVM CE 25-dev+37.1 (build 25+37-jvmci-b02)
Substrate VM GraalVM CE 25-dev+37.1 (build 25+37, serial gc)
```

This is neither GraalVM 25, nor GraalVM 26. It's graal/master with labs jdk 25.

Therefore, the allow-listing done in #364 no longer matches for such builds, but it should.

jerboaa:
Note that we don't know yet what the next version of graal/master will be.

PR #370, opened 2025-09-26
Allow list new warnings message on JDK 25+

jerboaa:
Previously this was done on JDK 26+, but with the recent reversion of graal/master being JDK 25-based again we don't get the allow-listed warnings matched anymore.

Instead of doing a new version parsing scheme (again!) we just match the version 25+ as it should do no harm to match this on Mandrel 25 as well.

Closes: #369


jerboaa:
The one CI failure seems unrelated.

PR #371, opened 2025-09-28
Additional JDK25 and Quarkus 3.28.1 whitelists

Karm:


Karm:
@jerboaa Some more whitelists; GHA does not show it because GHA does not run the full suite.

testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
```diff
@@ -300,6 +300,16 @@ public Pattern[] get(boolean inContainer) {
...
+                    p.add(Pattern.compile(".*JDBC fetch size: undefined/unknown.*"));
+                    p.add(Pattern.compile(".*Pool: undefined/unknown.*"));
+                    p.add(Pattern.compile(".*catalog/schema: unknown/unknown.*"));
+                }
+            }
+            if ((UsedVersion.getVersion(inContainer).compareTo(Version.create(25, 0, 0)) >= 0)) {
+                // https://github.com/quarkusio/quarkus/issues/47769
+                p.add(Pattern.compile(".*to use the thread-local-reset capability on Java 24 or later.*"));
+                p.add(Pattern.compile(".*Could not initialize class org.jboss.threads.JDKSpecific.*"));
             }
```
jerboaa, line 313: Where do we see those? A better fix would be to run `native-image` with `-J--add-opens=java.base/java.lang=ALL-UNNAMED` since we know it's needed for a newly generated quarkus app. This avoids not seeing them when we actually don't expect it.

Karm: >  The current approach in this patch disables the warning for all apps in the TS when run on Mandrel 25+.

Not true, it affects only this app: QUARKUS_MP_ORM_DBS_AWT

So it's really a matter of either adding the `-J--add-opens` to the command line or whitelisting it here.

I am O.K. with adding it to the app's command line though, so I'll refactor it...

jerboaa: > > The current approach in this patch disables the warning for all apps in the TS when run on Mandrel 25+.
> 
> Not true, it affects only this app: QUARKUS_MP_ORM_DBS_AWT

Ah, right. My bad.

Karm:
@jerboaa So the topic of opening java.base to ALL-UNNAMED and allowing native access to ALL-UNNAMED is resolved with this PR in both pertinent areas:

 * Maven itself via `.mvn` properties
 * native-image via `-J` flag, sometimes as a command line arg, sometimes in application.properties

PR #384, opened 2025-11-26
Prepare the arrival of UBI 10 and do a lot of Spring cleaning

gsmet:
I wasn't able to test everything as some testing requires sudo for whatever reason so let's see how CI likes it.

This is to prepare the arrival of UBI 10 images: https://github.com/quarkusio/quarkus-images/pull/313. I started by wanting some test coverage for UBI 10 and this is how it ended up :).

I took Quarkus 3.20 as the new base, which allowed to simplify a lot of stuff.

Note that commits are semantic and should be reviewed one by one.

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
@@ -82,6 +82,10 @@ public enum WhitelistLogLines {
             p.add(Pattern.compile(".*Minimum pool size: undefined/unknown.*"));
             p.add(Pattern.compile(".*Isolation level: <unknown>.*"));
             p.add(Pattern.compile(".*Maximum pool size: undefined/unknown.*"));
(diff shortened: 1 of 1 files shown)


Karm:
Hello @gsmet, thank you for visiting us at the Mandrel Integration Testsuite venue :-) 

I went over the commits and we do not need these changes at this time except this one:

 *  [74d0fbe](https://github.com/Karm/mandrel-integration-tests/pull/384/commits/74d0fbe781bab521a91ea4240dcb85f9d2810e30) Provide a list of runtimes instead of excluding some 
 
I like that change a lot and we would like to have it and backport it to the branch https://github.com/Karm/mandrel-integration-tests/tree/testing-more-runtime-images that is used to run tests over at https://github.com/quarkusio/quarkus-images 
If you don't mind, I would drop your other commits and integrate this one.

Karm:
Hmm, while testing it I remembered why I did this "such and such is excluded" in the first place.
It was to have a convenient way to print these black squares saying that yes, we know about that runtime, but it was intentionally skipped and why:

image

The current approach seems to make this cumbersome...

PR #385, opened 2025-11-26
Added three error lines into the whitelist

patrikcerbak:
Hi,
On newer Quarkus versions, these three new error lines popped up, adding them into the whitelist would fix `testQuarkusMPOrmAwtContainer`.

@Karm, we discussed this on Slack yesterday when I tried running it locally, but the test also popped up in infra now with this failure.

What do you think?

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
@@ -82,6 +82,11 @@ public enum WhitelistLogLines {
             p.add(Pattern.compile(".*Minimum pool size: undefined/unknown.*"));
             p.add(Pattern.compile(".*Isolation level: <unknown>.*"));
             p.add(Pattern.compile(".*Maximum pool size: undefined/unknown.*"));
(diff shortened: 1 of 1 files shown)


testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
```diff
@@ -82,6 +82,11 @@ public Pattern[] get(boolean inContainer) {
             p.add(Pattern.compile(".*Minimum pool size: undefined/unknown.*"));
             p.add(Pattern.compile(".*Isolation level: <unknown>.*"));
             p.add(Pattern.compile(".*Maximum pool size: undefined/unknown.*"));
+            if (QUARKUS_VERSION.compareTo(new QuarkusVersion("3.27.0")) >= 0) {
+                p.add(Pattern.compile(".*JDBC fetch size: undefined/unknown.*"));
+                p.add(Pattern.compile(".*Pool: undefined/unknown.*"));
+                p.add(Pattern.compile(".*Default catalog/schema: unknown/unknown.*"));
+            }
```
gsmet: @yrodiere do you know why these are suddenly popping up? It's not happening in all cases AFAICS.

PR #388, opened 2025-12-10
Added a Podman warning and error into the whitelist

patrikcerbak:
Hi, I added two harmless Podman logs into the error whitelist. They shouldn't affect any test results.

Links to discussions about the logs:
https://github.com/containers/podman/discussions/19920
https://github.com/containers/podman/issues/16217

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
@@ -338,6 +338,8 @@ public enum WhitelistLogLines {
             }
             // harmless Podman noise, does not affect the test outcome (https://github.com/containers/podman/discussions/19920)
             p.add(Pattern.compile(".*time=\".*\" level=warning msg=\"archive: skipping.*"));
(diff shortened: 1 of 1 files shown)


Karm:
@patrikcerbak 

The 3.27.1 gdb error is tracked here: https://github.com/graalvm/mandrel/issues/919, so no-op in this PR.

The 3.8.6.1 error: Isn't is just Jaeger not running? It should be running on Linux though....   Race condition? If it shows in HotSpot mode too I'd say whitelist it.

```
testsuite/target/archived-logs/org.graalvm.tests.integration.PerfCheckTest/testQuarkusMPOrmAwtLocal/build-and-run.log and check these offending 1 lines: 
2025-12-10 14:46:38,432 SEVERE [io.qua.ope.run.exp.otl.VertxGrpcExporter] (executor-thread-3) Failed to export spans. The request could not be executed. Full error message: Client is closed [Error Occurred After Shutdown] ==> expected: <true> but was: <false>
```

Karm:
`Recursive internal problem` is https://github.com/graalvm/mandrel/issues/919

Karm (approved):

PR #389, opened 2025-12-10
Made the new builder image compatible with tests

patrikcerbak:
Hi, I added the new builder image name into `isBuilderImageIncompatible()` method, so it 's compatible with the tests.

What do you think? Is it too specific?

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
@@ -1380,7 +1380,7 @@ public class AppReproducersTest {
      */
     public static boolean isBuilderImageIncompatible(String base) {
         // it's UBI 8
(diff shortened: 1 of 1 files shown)


testsuite/src/it/java/org/graalvm/tests/integration/AppReproducersTest.java
```diff
@@ -1380,7 +1380,7 @@ private static Map<String, String> getSwitches(Apps app) {
      */
     public static boolean isBuilderImageIncompatible(String base) {
         // it's UBI 8
-        if (BUILDER_IMAGE.contains("/ubi-")) {
+        if (BUILDER_IMAGE.contains("/ubi-") || BUILDER_IMAGE.contains("/quarkus-mandrel-for-jdk-21-rhel8")) {
```
Karm: @patrikcerbak Why not just `rhel8` ?

Karm (approved):
@patrikcerbak THX. It's fine. The test error is the gdb one: https://github.com/graalvm/mandrel/issues/919

PR #405, opened 2026-04-30
Allow warning for using 23.1 instead of 25.0

zakkak:
Prevents failures when using 23.1

Warning introduced with https://github.com/quarkusio/quarkus/pull/53820

Failure seen in https://github.com/graalvm/mandrel/actions/runs/25052464184/job/73389241996?pr=961

### DIFF
--- a/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
+++ b/testsuite/src/it/java/org/graalvm/tests/integration/utils/WhitelistLogLines.java
@@ -61,6 +61,8 @@ public enum WhitelistLogLines {
             p.add(Pattern.compile(".*Warning: Please re-evaluate whether any experimental option is required, and either remove or unlock it\\..*"));
             // Quarkus main 2024-03-21 deprecates this maven plugin directive
             p.add(Pattern.compile(".*Configuration property 'quarkus.package.type' has been deprecated.*"));
+            // Allow the quarkus warning about older Mandrel/GraalVM versions
+            p.add(Pattern.compile(".*You are using an older version of GraalVM or Mandrel : 23\\.1.* Quarkus currently supports 25\\.0.* Please upgrade to this version\\..*"));
             // Microdnf complaining, benign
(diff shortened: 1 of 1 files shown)


jerboaa (approved):
LGTM

zakkak:
Failure 1 is:

```
Error:  Failures: 
Error:    DebugSymbolsTest.debugSymbolsQuarkusContainer:410 GDB session did not start well. Check the names, paths... Content was: thread-group-added,id="i1"GNU gdb (Red Hat Enterprise Linux) 16.3-2.el9
```

Failure 2 is:

```
Error:  Failures: 
Error:    RuntimesSmokeTest.quarkusFullMicroProfile:227->testRuntime:172 build-and-run.log log should not contain error or warning lines that are not whitelisted. See D:\a\mandrel-integration-tests\mandrel-integration-tests\ts\testsuite\target\archived-logs\org.graalvm.tests.integration.RuntimesSmokeTest\quarkusFullMicroProfile\build-and-run.log and check these offending 1 lines: 
Warning:  [io.quarkus.runtime.util.ContainerRuntimeUtil] Command "docker" exited with error code 1. Rootless container runtime detection might not be reliable. ==> expected: <true> but was: <false>
```

Both seam unrelated (and I wonder why we are still using such old versions for testing these).

Karm:
> Both seam unrelated 

@zakkak 

* GDB:  https://github.com/graalvm/mandrel/issues/919

* `docker` -  Quarkus is looking for it, it's removed fro the runner, warning is emitted.  Should be whitelisted hten

> (and I wonder why we are still using such old versions for testing these).

The oldest one for this TS should be Quarkus 3 now [1], so I believe we can begin to retire Quarkus 2 now.  I'll check after CPU, have a chat with @patrikcerbak. Means that the Q apps that are written for Q 2 and patched for Q 3 would be Q 3 only now. 

[1]  https://code.quarkus.redhat.com/?S=com.redhat.quarkus.platform%3A3.2

